PR #7, opened 2019-05-09
functionality to init IG from scratch v2

roelandluykx:
Proposal of additional functionality to init a new IG added.


grahamegrieve:
@jamesagnew can you please review the Maven related changes here

grahamegrieve:
I don't understand these changes. The specified folder to ini has content copied from adHocTmpDir  - but what's in there?

jamesagnew:
@roelandluykx Can you comment on the intent behind the two pom.xml changes in this change?

As far as I can see they are just reformatting. Maybe I've missed a subtle change though? If it's just formatting, can you rebase these against the current master so that they don't show up as changes here?

roelandluykx:
@jamesagnew 
It is not only formattins IG.
After checking out i could not generate the jar file to run an ig without missing classes. so i analyzed the maven and recognized, that it is not setup according to how to make multimodule maven project.
the project is set up as an eclipse project and not as an maven project. there are classpasses and project dependencies linked by eclipse. in a pure maven project, no .project, .classpath and no .settings files and folders should be checked in into git.
so i reorganized the pom to have the dependencyManagement in the main pom and the dependencies to be used in the sub module poms. there the dependencies are included as the has to be and the indirect dependencies are managed by maven.

roelandluykx:
@grahamegrieve i added only the init functionality to run the rest of the code as it was implemented.
the content of the adHocTmpDir  is done org.hl7.fhir.igtools.publisher.Publisher.copyDefaultTemplate() where a zip file from the selected package i.e ~/.fhir/packages/hl7.fhir.core#4.0.0/other/ig-template.zip.
the config file is generated by the code org.hl7.fhir.igtools.publisher.Publisher.buildConfigFile().
thats what i have analyzed. based on this the init method is done.
in a further version more questions could be asked to fill in the config directly with the poject stuff.

jamesagnew:
@roelandluykx While I agree with the general principal of not checking in IJ/Eclipse project files into Git (we don't do this in the core HAPI repo for example) this repo/app is used in Eclipse by some people don't have long histories of troubleshooting eclipse. I think I'd prefer to keep them in in this case.

Are the Maven changes strictly necessary in order to implement this new feature? If not, it probably makes sense to reduce the scope of the PR to just the changes that are required.

roelandluykx:
@jamesagnew 
No you are right, it does not have to do with the new functionality i added in the code.
But i was not able to build locally a jar file and use it to init a new or build an existing IG without NPE's or CNE... so i was not able to test the ig-publisher loacally.

grahamegrieve:
we have several people struggling with debugging this now. Apparently what works for me (which is still unreliable) is not reproducible for others...

lmckenzi:
This appears to have been abandoned and is very old.  If still relevant, please re-submit against the current code base.

PR #16, opened 2019-06-21
Table desc fixes

brianpos:
When generating the table segments for SearchParameters, OperationDefinitions and CapabilityStatements the Description column does not extract the data from the resource (and instead uses the name)
Also, fix the spelling error on the readme page , Lloyds name isn't _Llod_

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,4 +19,4 @@ mvn -Dmaven.test.skip install
 
 == Maintenance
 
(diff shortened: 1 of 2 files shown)


grahamegrieve:
That code is crap. I abandoned this and fixed it properly

grahamegrieve:
uh, that is, the original code, not the PR

brianpos:
Oh, then you'll want to take another look at the other pull request, and maybe cherry pick bits of it.
(otherwise I'm happy to make those changes)

PR #18, opened 2019-06-25
Improve rendering of tables and include an examples table as an optional output

brianpos:
For many resource types there isn't a description column and these will then just default in the name of the resource again (or the Id). Instead of doing this, leverage the name that is provided by the Implementation Guide
Also:
* close the file stream when processing the transaction log output
* create a new output snippit - table format of the examples list
* the title used when a fetched resource doesn't have one was changed from `type(space)id` to a normal resource reference `type/id`

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,4 +19,4 @@ mvn -Dmaven.test.skip install
 
 == Maintenance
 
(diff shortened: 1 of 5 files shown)


brianpos:
This has now been replaced by PR#19

PR #21, opened 2019-07-09
generate constraint snapshots on logical models

oliveregger:
For logical models which have derivationtype constraint the snapshot was done the same as for specialised (union), now the behaviour is changed that the same elements are constrained in the differential as in other StructureDefinitions. See also https://github.com/HL7/cda-core-2.0/pull/11.


grahamegrieve:
Took to long to get to this, so I just made the code change directly

PR #23, opened 2019-08-01
Remove extra closing paragraph from injected status box text

lmckenzi:


grahamegrieve:
turned into a conflict so I fixed it directly

PR #24, opened 2019-08-04
Trap isExample() NullPointerException to fix build.

rhausam:
Trap the NullPointerException thrown by the isExample() function in order to fix the build.  There likely is a better and more permanent fix to be made, but this allows the build to complete.


rhausam:
This seems to be an issue when using Lloyd's framework and possibly not with Eric's.  I have run into the issue with the LIVD IG and Jose Teixiera has it one of his, and both use Lloyd's framework.  The IPS IG is using Eric's framework and it doesn't have the issue.

If the pull request is accepted (in lieu of another solution), obviously the console error message can be improved.

rhausam:
Grahame made a slightly different fix and the issue is resolved.

PR #31, opened 2019-09-05
Don't treat inter-version extensions as errors

lmckenzi:


grahamegrieve:
This regex looks very permissive compared to the scope of the problem?

lmckenzi:
It's already in the scope of "starts with "http://hl7.org/fhir/StructureDefinition", so we're really just checking the format of the id.

grahamegrieve:
It allows lots of non-valid extensions. In fact, it won't complain about any HL7-ish extension, whether they're valid or not? that's not good

lmckenzi:
Thoughts on a better process?

grahamegrieve:
abandoned as I implemented an actual correct solution to this requirement

PR #37, opened 2019-10-18
use snapshot generation for logical/specialization

oliveregger:
this pull requests updates the ig publisher to use for logical models and structure definitions with derivation type specialization the snapshot generation functionality ([see merged pull request which introduced this functionality in org.hl7.fhir.core](https://github.com/hapifhir/org.hl7.fhir.core/pull/90))


grahamegrieve:
I don't understand this change - didn't it do that already?

oliveregger:
It has been done for the general snapshot generation in ProfileUtilities, however the IG Publisher uses an additional function utils.populateLogicalSnapshot(sd) which just copies the element. An alternative would be to change the functionality in populateLogicalSnapshot, would you prefer that?

PR #49, opened 2019-12-21
Writing lower-case file paths to the preProcessInfo hash map and retr…

seanmcilvenna:
…ieving them as lower case since paths are being found with a mix of upper-case and lower-case representations.


grahamegrieve:
@lmckenzi please check

lmckenzi:
This pull request fixes the issue - but do we want to confirm that the passed in case is correct for other reasons or allow it to propoagate?

seanmcilvenna:
My suggestion is to start with this and make that determination later. Would like the ig publisher to work in the ToF dev server... and right now it doesnt.

PR #53, opened 2020-01-20
Code was erroneously checking for absolute file path rather than abso…

lmckenzi:
…lute URL


grahamegrieve:
I've already made this change, so not merging to avoid a merge conflict elsewhere

PR #54, opened 2020-01-21
Fix issue where example isn't loaded when creating references to asso…

lmckenzi:
…ciated profile


grahamegrieve:
made changes to the code directly after discussion around this bug.

PR #55, opened 2020-01-28
Adding logic to make sure that there are differential elements in a p…

seanmcilvenna:
…rofile before calling generateSnapshot(). An error is reported in QA if there are no elements in the differential, as opposed to just throwing an exception.


grahamegrieve:
Solve the problem a different way instead of using this PR after discussion with Sean

PR #60, opened 2020-02-16
When pointing to a local template that inherited from a remote packag…

lmckenzi:
…e template where both had 'config' files from a Windows machine using Eclipse, I ran into a problem where the NPM package manager treated the folders as using the windows separator "\" but the local package used the windows separator "/".  I changed the code to check for both.  (It may be that we should be changing code somewhere else, but I'm not sure where.)


grahamegrieve:
This was solved a different way

Issue #62, opened 2020-03-12
Support for non-english language


peturv:
I've been working with an IG where a requirement is delivery in a non-english language.
One of the localisation issues is text snippets generated by the publisher tool.

Am submitting the code I have been using to handle language localisation in the publisher so far, hoping it will be of use.

lmckenzi:
Will leave #63 active.  The general question of how translations will be handled will be dealt with as part of the WHO-funded project over the coming year.

PR #63, opened 2020-03-12
Added bundle-based language support to TranslationImplementation.

peturv:
Support for harvesting keys to console by passing "harvest" as language.
Message bundles based on harvested keys.

Language parameter passed in as a param in IG control file. Correct BCP47 format expected.

Associated with issue #62


peturv:
N.b. The Swedish translation file is just a proof of concept. Those are not validated translations.

org.hl7.fhir.publisher.core/src/main/resources/Messages_sv_SE.properties
```diff
@@ -0,0 +1,66 @@
...
+sd.tx.path=Sökväg
+sd.tx.terminology_bindings_differential=Terminologibindningar (Differential)
+sd.tx.terminology_bindings=Terminologibindningar
+sd.tx.valueset_code=ValueSet / Code
+sd.tx.valueset=ValueSet
+vs.summary.defining_url=Definierande URL
+vs.summary.name=Namn
+vs.summary.publisher=Utgivare
+vs.summary.source_resource=Källresurs
+vs.summary.title=Titel
```
markiantorno: If this file is just for testing, and the translations are not correct, we shouldn't be checking it into the main branch. It would probably be good idea to keep test code in branches. 
Please remove.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java
```diff
@@ -1386,6 +1375,33 @@ else if (p.getValue().equals("no-broken-links"))
...
+                  .map(String::trim)
+                  .filter(s -> s.length() > 0)
+                  .findFirst()
+                  .orElseGet(() -> {
+                    System.out.println("Language parameter defined, but no valid languages found");
+                    return null;
+                  });
+          ((TranslationImplementation)this.translator).setLang(firstLanguage);
+        }
+        */
```
markiantorno: I"m actively trying to reduce the amount of commented out code in the libraries right now. If a feature isn't ready to be checked in, let's try and keep it in a branch. Using the master branch to store wip code really isn't ideal.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -21,40 +21,56 @@
...
+  private ResourceBundle bundle = ResourceBundle.getBundle("Messages");
+  private boolean harvest = false;
+
   public String getLang() {
     return lang;
   }
 
   public void setLang(String lang) {
-    this.lang = lang;
+    if("harvest".equals(lang)) {
```
markiantorno: Put any magic numbers or strings into final variables with commented explanations of what their purpose is please.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -21,40 +21,56 @@
...
   // marker is used to see what content is not being translated (debugging)
   private static final String MARKER1 = ""; // "^";
   private static final String MARKER2 = ""; // "^^";
-  
+  private Map<String, String> keys = new HashMap<>();
   // -- configuration -------------------------------------------------------
   
   private String lang; 
-  
+  private String nonAlpha = "[^a-zA-Z0-9_]+";
```
markiantorno: This field should be final unless you're planning on changing it? Also, please use proper naming conventions for the variables, like NON_ALPHA for final constants.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -21,40 +21,56 @@
...
+  private String noEndDash = "_$";
```
markiantorno: This field should be final unless you're planning on changing it? Also, please use proper naming conventions for the variables, like NO_DASH_END for final constants.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -21,40 +21,56 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.*;
+
 public class TranslationImplementation implements TranslationServices {
 
   // marker is used to see what content is not being translated (debugging)
   private static final String MARKER1 = ""; // "^";
   private static final String MARKER2 = ""; // "^^";
-  
+  private Map<String, String> keys = new HashMap<>();
```
markiantorno: Preface all member fields with m, so this could be mKeys. Helps differentiate when debugging methods.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -21,40 +21,56 @@
...
   // -- configuration -------------------------------------------------------
   
   private String lang; 
-  
+  private String nonAlpha = "[^a-zA-Z0-9_]+";
+  private String noEndDash = "_$";
+  private Locale locale = Locale.getDefault();
+
+  private ResourceBundle bundle = ResourceBundle.getBundle("Messages");
+  private boolean harvest = false;
```
markiantorno: Again, for all these variables, please use proper naming conventions (ie. preface with 'm' for member vars, 'mIs' for boolean fields)

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -21,40 +21,56 @@
...
   }
 
   public void setLang(String lang) {
-    this.lang = lang;
+    if("harvest".equals(lang)) {
+      this.harvest = true;
+    } else {
+      this.lang = lang;
+      this.locale = Locale.forLanguageTag(lang);
+      bundle = ResourceBundle.getBundle("Messages", this.locale);
```
markiantorno: Put any magic numbers or strings into final variables with commented explanations of what their purpose is please.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -21,40 +21,56 @@
...
+  private Locale locale = Locale.getDefault();
+
+  private ResourceBundle bundle = ResourceBundle.getBundle("Messages");
+  private boolean harvest = false;
+
   public String getLang() {
     return lang;
   }
 
   public void setLang(String lang) {
```
markiantorno: I'm really not a fan of taking a method that for all intents and purposes appears to be a normal accessor method, and then packing additional functionality under the hood.
If you do take this approach, so that it is no longer is as simple as setting the language field as it name would suggest, please add a javadoc detailing what this method does. It is no longer straightforward.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -21,40 +21,56 @@
...
+      this.lang = lang;
+      this.locale = Locale.forLanguageTag(lang);
+      bundle = ResourceBundle.getBundle("Messages", this.locale);
+    }
   }
 
   // -- services -------------------------------------------------------
-  
+
   public String translate(String context, String string) {
```
markiantorno: javadoc here please

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -21,40 +21,56 @@
...
-    return string == null ? null : MARKER2+string+MARKER2;
+    if(string == null) return null;
+    String key = generateKey(context, string);
+    return harvest ? string : resolve(key);
   }
 
   public String translate(String context, String string, Object... args) {
```
markiantorno: javadoc here please

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -104,5 +120,38 @@ public String translateAndFormat(String contest, String lang, String string2, Ob
     return null;
   }
 
+  public void printKeys(){
```
markiantorno: javadocs here please

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -104,5 +120,38 @@ public String translateAndFormat(String contest, String lang, String string2, Ob
...
+    if(harvest) {
+      System.out.println("Printing bundle keys:");
+      keys.forEach((key, value) -> System.out.println(String.format("%s=%s", key, value)));
+    } else {
+      System.out.println("Not harvesting bundle keys. Nothing printed");
+    }
+  }
+
+  private String generateKey(String context, String path) {
```
markiantorno: javadoc for this method please

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/TranslationImplementation.java
```diff
@@ -104,5 +120,38 @@ public String translateAndFormat(String contest, String lang, String string2, Ob
...
+
+    return key;
+  }
 
+  // TODO: Handle by custom ResourceBundle.Control instead. This is just a workaround
+  // Should language properties be partitioned into bundles by context instead of current prefixing?
+  private String resolve(String key) {
+    return new String(bundle.getString(key).getBytes(StandardCharsets.ISO_8859_1), StandardCharsets.UTF_8);
+  }
 }
```
markiantorno: I know that testing for this library is where it needs to be currently, but I would prefer if you added tests to cover the new functionality in this file you have added. Even if it's just testing the key management method and the calls the the modified methods that now check if "harvest" is set. I'm working on getting some better CICD pipelines set up for a bunch of the core libraries, including this one and this is the kind of approach that would be good to take moving forward.

If not too complicated, it would be great to test the actual fetching of pairs using the 'harvest' variable.

peturv:
@markiantorno Thanks for the input. I wasn't sure if this was of interest to you, rather wanted to share my how I am doing this. I'll gladly polish this up a bit according to your input.

lmckenzi:
No progress in 2 years.  We're now pursuing a different approach to translations based on WHO funding

PR #64, opened 2020-03-26
Restore rendering the supplied ValueSet.text for the cld.

rhausam:
This is needed for rendering large extensionally defined value sets where it isn't reasonable to enumerate the contents of the definition as is done in the default generated ValueSet rendering.  Doing this also greatly improves the IG build performance with the large extensionally defined value sets (which otherwise can become extremely slow and nearly unusable).


rhausam:
This pull request is superseded by Grahame's commit 63127dfea42dc31f624229a59366bea9c410a0a7.

PR #98, opened 2020-05-20
Fix issue with generated IG in output folder not including all parame…

lmckenzi:
…ters


grahamegrieve:
Please explain this change. I don't understand why it's a good idea

lmckenzi:
Eric had raised an issue that all of the parameters weren't showing up in the published IG.  Different templates inject different parameters.  Presumably knowing what actually drove publication is deemed to be relevant.

grahamegrieve:
My question wasn't about the intent of the change, but why the actual code changes proposed in the PR are a good idea. Why is it safe not to load the element model?

lmckenzi:
Because what it loads is the original IG, not the one with the modifications from the template.

PR #104, opened 2020-05-28
Fix the location property in the US-core validation message so it act…

lmckenzi:
…ually identifies the profile with the issue (when there are multiple profiles in a spreadsheet, StructureDefnition.baseDefinition doesn't help).


grahamegrieve:
This is not right - the path must be a valid FHIRPath statement that identifies where in the resource the problem is

lmckenzi:
The issue is that it the resource is in a spreadsheet or a Bundle, there's no indication of what profile has the problem.  The current path provides no useful information.  There are other issues reported that put the canonical URL in the location field, so I was just aligning with those.

grahamegrieve:
Where's the other issues?

lmckenzi:
image

grahamegrieve:
ok thanks. fixed those so they are no longer wrong.

PR #105, opened 2020-05-31
Expose the source file in _data/resources.json so that it can be expo…

lmckenzi:
…sed in templates.  (UTG needs this to support editing & pull requests and other IGs might want the same.)


grahamegrieve:
I think that f.getRelativePath() would be safer than using substring()

lmckenzi:
relativePath is null - at least at the time this call is made.

grahamegrieve:
hmm handled differently then.

PR #127, opened 2020-07-08
Version number fix

markiantorno:


markiantorno:
I have run this against the latest US core, and verified that the 'SEARCHPARAMETER_EXP_WRONG' no longer appears in the qa.html results.

PR #219, opened 2020-12-23
Update Publisher.java

markkramerus:
We have seen cases where 1 minute isn't sufficient for completing SUSHI. One case was 3 minutes, 20 seconds. I suggest changing to 5 minutes.


grahamegrieve:
well, ok, though you not about fshtimeout in fsh.ini, right?

grahamegrieve (approved):
ok

PR #249, opened 2021-04-04
Issue#103 - Fixed NRE in Bundle validation for entries with no fullUrl

sliver007:
Addresses Issue# 103 - Fixes a nullReferenceException that results from validation of a Bundle with entries that have no fullUrl.


grahamegrieve (commented):
ok

PR #257, opened 2021-05-03
No history publish

lmckenzi:
Ensure we don't publish Provenance and Provenance bundles that represent history of other resources as resource pages


grahamegrieve (approved):
ok

PR #281, opened 2021-06-08
Don't raise an error if an element is identified as an example of an …

lmckenzi:
…inherited profile rather than one defined in the current resource.


grahamegrieve (approved):
ok

Issue #285, opened 2021-06-17
Improve support for directories and GitHub URLs as template base


gabriel0316:
Within template's `package.json` it is possible to specify that the current template is based on another one, e.g.:

    {
      "name": "some.template",
      "type": "fhir.template",
      "license": "CC0-1.0",
      "description": "Some sample IG template based on another one",
      "author": "someone",
      "canonical": "http://somesitesomewhere.com/templates/some.template",
      "base": "fhir.base.template",
      "dependencies": {
        "fhir.base.template": "current"
      },
      "version": "0.0.1"
    }

If `base` is specified `dependencies` is **required** to contain an entry which corresponds to `base` (see above).

In theory it is also possible to specify a directory or a GitHub repository.

However, the following line in [`TemplateManager.java`](https://github.com/HL7/fhir-ig-publisher/blob/master/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/templates/TemplateManager.java#L108) raises some problems:

    installTemplate(baseTemplate+"#"+ver, rootFolder, templateDir, scriptIds, loadedIds, level + 1);

`baseTemplate` contains the value from `base` and `ver` contains the value from `dependencies."fhir.base.template"`.

If I now want to specify a directory (e.g. `./myTemplate`) or a GitHub repository (e.g. `https://github.com/HL7/ig-template-fhir`) and the corresponding `dependency` is `"current"` this would lead to the fact, that `installTemplate` is being invoked with the following values for the first parameter:

- `./myTemplate#current` in the case of the directory
  - As a **workaround** the directory of the template would have to be renamed to `myTemplate#current` in order to be retrieved correctly.
- `https://github.com/HL7/ig-template-fhir#current` in the case of the GitHub repository
  - As a result the retrieval of the repository's template does not work because in `TemplateManager#zipUrl()` (Line [281](https://github.com/HL7/fhir-ig-publisher/blob/master/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/templates/TemplateManager.java#L281)) "archive" and "master.zip" will be added resulting in `https://github.com/HL7/ig-template-fhir#current/archive/master.zip`

Probably the solution to this is quite simple. I also have found an easy workaround for the second case. However, I'm not quite sure about possible side effects - that's the reason why I've created this issue.

I do also think that for a directory `dependencies` should not be required as there can only be one version in a directory. 

[...]

PR #300, opened 2021-07-23
Add quickfix for GitHub and directory based templates

gabriel0316:
See https://github.com/HL7/fhir-ig-publisher/issues/285


grahamegrieve (commented):
can you explain these two changes?

gabriel0316:
Reason for this change is explained in #285 

However, as it is also possible to publish templates to build.fhir.org using the auto-ig-builder the changes above are more or less obsolete.

gabriel0316:
BTW - including the commit regarding `.vscode/launch.json` was a mistake and should be ignored.

PR #309, opened 2021-09-02
Clarify "not used in this Implementation Guide" is about profiles

jmandel:
See https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Extension.20examples


grahamegrieve (approved):
ok

PR #319, opened 2021-09-29
work around for bad binding descriptions in R4, and allow editor to s…

grahamegrieve:
…uppress validation or generation (for speed)


grahamegrieve:
@markiantorno for your review

PR #339, opened 2021-11-18
add -noprompt arg for release updater

### LINKED ISSUE #336: Allow the release update process to skip the "Proceed Y/N" prompt
Currently the release update always asks for a Y/N answer (apparently unless there is a -filter in the command line?).

It would be good to be able to use the release update process in an automated way - not for fully automating the release, but for allowing a fast iteration when finetuning a Release Candidate, for example.

Can we add a command line option that skips the Y/N prompt?

costateixeira:
Fixes #336 by adding a -noprompt command line argument


grahamegrieve (approved):
ok

PR #363, opened 2021-12-22
Clarify jurisdiction message for FSH

vadi2:
It wasn't immediately obvious to a FSH beginner that the line should go into the yaml config.


grahamegrieve (approved):
ok

PR #370, opened 2022-01-16
Changes needed to support revised artifact table rendering

lmckenzi:


grahamegrieve:
what does propagateStatus() do, and where is it documented?

lmckenzi:
It migrates standards status and FMM through dependencies unless overridden.  Where do you want it to be documented?

grahamegrieve:
Well, the parameters should be documented here: https://confluence.hl7.org/display/FHIR/Implementation+Guide+Parameters. I suspect that the way standards status migrates should also be documented on that page, and also in the code

lmckenzi:
Replaced by #386

PR #371, opened 2022-01-16
refactor how web site layout naming rules work

grahamegrieve:


costateixeira (changes requested):
[org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/utils/PublicationProcess.java, Line 219
](https://github.com/HL7/fhir-ig-publisher/pull/371/files#diff-ccc6b4e460052b5fd03d8984f50b8d8b82bc0be544295b962203c8c622bd16a1L219)

Mentions DK but it's BE. Suggest to 
change to 
" does not match the required canonical of http://hl7belgium.org/profiles/fhir/ or http://ehealth.fgov.be/standards/fhir/"

PR #386, opened 2022-02-02
Fmm propagation and contained support

lmckenzi:
Replaces #370, with some bug fixes and also fixes to support for contained resource rendering


grahamegrieve (approved):
ok

PR #400, opened 2022-02-17
Sdc r4 example issue

lmckenzi:
npm package list wasn't being updated after importing a new package - which is what happens every time on the CI build.  As a result, example resolution wasn't checking against those imported packages.  Moved the point of adding the package to the list post-import.  (If there's a reason why the adding to the list was happening post-import, then need to find some other way of ensuring that examples are de-referenced properly by the validator.)


lmckenzi:
Contained commits that weren't intended

PR #405, opened 2022-02-24
fix one more issue with Belgian URLs

costateixeira:
Sorry, I left out the www in the previous PR. I'm told these are the correct URLs


grahamegrieve (approved):
ok

PR #410, opened 2022-03-04
Stu3 template fixes

lmckenzi:


grahamegrieve (commented):
uh, why? I don't like these changes at all

lmckenzi:
The relevant value sets don't exist for all versions of FHIR, so defining them inline was the best way to make this work.  Why do you not like the change?

grahamegrieve:
I generally try to avoid defining them in code. Why aren't they in THO? Or should they be in pubpack?

lmckenzi:
Is THO used with STU3 and earlier releases?
In any event, it only has value sets for 2-char and 3-char country, not numeric.  Also doesn't have the state codes.

grahamegrieve:
THO is loaded for R3 but not R2, I think. You can move them out of the code at some stage

PR #425, opened 2022-04-11
When a watchdog process times-out, log a timeout.

rob-reynolds:
ISSUE: https://github.com/HL7/fhir-ig-publisher/issues/427

The following message is very confusing in the case of a watchdog timeout:
"Note: Check that [launched process] is installed correctly."

Since the actual case is known, log it.


grahamegrieve (approved):
ok

PR #424, opened 2022-04-11
Update all fasterxml jackson versions to match the jackson-databind v…

rob-reynolds:
…ersion

ISSUE: https://github.com/HL7/fhir-ig-publisher/issues/426

When the dependabot updated the com.fasterxml.jackson.core jackson-databind dependency (on 3/30, sha d2109d5acbf5aa974b0f98651c0fa53214bd3c88) , it caused an error in the jackson-core dependency, in some cases:

Translating CQL source in folder ./input/cql (01:45.0556)
Exception in thread "main" java.lang.NoClassDefFoundError: com/fasterxml/jackson/core/util/JacksonFeature
at com.fasterxml.jackson.databind.ObjectMapper.<init>(ObjectMapper.java:655)
at com.fasterxml.jackson.databind.ObjectMapper.<init>(ObjectMapper.java:558)
at org.cqframework.cql.cql2elm.CqlTranslatorOptionsMapper.<clinit>(CqlTranslatorOptionsMapper.java:10)
at org.hl7.fhir.igtools.publisher.CqlSubSystem.getTranslatorOptions(CqlSubSystem.java:279)
at org.hl7.fhir.igtools.publisher.CqlSubSystem.translateFolder(CqlSubSystem.java:294)
at org.hl7.fhir.igtools.publisher.CqlSubSystem.execute(CqlSubSystem.java:223)
at org.hl7.fhir.igtools.publisher.Publisher.load(Publisher.java:3928)
at org.hl7.fhir.igtools.publisher.Publisher.createIg(Publisher.java:1002)
at org.hl7.fhir.igtools.publisher.Publisher.execute(Publisher.java:858)
at org.hl7.fhir.igtools.publisher.Publisher.main(Publisher.java:10049)
Caused by: java.lang.ClassNotFoundException: com.fasterxml.jackson.core.util.JacksonFeature
at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
... 10 more

The fix is to update all the other Jackson dependencies to the equivalent version.

PR #428, opened 2022-04-11
Feature configure Jekyll timeout

rob-reynolds:
ISSUE: https://github.com/HL7/fhir-ig-publisher/issues/429.

There are production content IGs that require Jekyll processing that exceeds the 5 minute default. Allow the value to be configured.

### DIFF
--- a/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java
+++ b/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java
@@ -595,7 +595,7 @@ public class Publisher implements IWorkerContext.ILoggingService, IReferenceReso
   private static final String REDIRECT_SOURCE = "<html>\r\n<head>\r\n<meta http-equiv=\"Refresh\" content=\"0; url=site/index.html\"/>\r\n</head>\r\n"+
        "<body>\r\n<p>See here: <a href=\"site/index.html\">this link</a>.</p>\r\n</body>\r\n</html>\r\n";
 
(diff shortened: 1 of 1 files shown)


rob-reynolds:
This may be too naïve an implementation to merge.

Not sure what the process is for adding things to the ImplementationGuideDefinitionParameterComponent type, whether that should be done first or if it should be 'trialed' like this until it warrants being added, etc.

Also, there are considerations for other ways configuration values can be read.

rob-reynolds:
Note: there are also 2 branches with separate PRs merged into this:
- https://github.com/HL7/fhir-ig-publisher/pull/424
- https://github.com/HL7/fhir-ig-publisher/pull/425

grahamegrieve (commented):
why rename JEKYLL_TIMEOUT to TOOL_TIMEOUT - it hasn't changed in function, it just gets less clear

rob-reynolds:
There's an abstraction in place to allow the specific "tool" to change: https://github.com/HL7/fhir-ig-publisher/blob/e15e191c5a114f0b8cd9d2cd4ad287323daf0273/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java#L6573.

The intent was to implement at that level.  

I see the point re JEKYLL_TIMEOUT.  Are you OK with the rest being at the higher level?

grahamegrieve:
It's a todo for me to get around to removing that abstraction - we've decided that we won't support anything other than Jekyll. And if I just possibly maybe did, there's no reason to think a different tool would even have a timeout, let alone the same one

rob-reynolds:
Does that look better?  

I also added configuration in the json version.  I can remove that if it's not appropriate.

rob-reynolds:
Deprecated

Issue #429, opened 2022-04-11
Feature: allow the timeout for Jekyll to be configured


rob-reynolds:
There are production content IGs that require Jekyll processing that exceeds the 5 minute default.  Allow the value to be configured.

Draft PR: https://github.com/HL7/fhir-ig-publisher/pull/428

lmckenzi:
@rob-reynolds is this now addressed?  You closed your PR, but it's not clear from the comments why.

lmckenzi:
Addressed.

PR #434, opened 2022-04-15
Add the ability to configure the timeout for jekyll

rob-reynolds:
ISSUE: https://github.com/HL7/fhir-ig-publisher/issues/429.

There are production content IGs that require Jekyll processing that exceeds the 5 minute default. Allow the value to be configured.


rob-reynolds:
Supersedes Feature configure Jekyll timeout #428.

Cleaned up commits.

grahamegrieve (approved):
ok

PR #459, opened 2022-07-04
add fhir.ch namingrulesprovider

oliveregger:
we have a swiss specific hl7 convention for package naming and publication, if we want to use -go-publish we have the need to add this specific naming rules: 

package name: ch.fhir.ig.[code]#version
canonical: http://fhir.ch/ig/[code]/version
realm: ch

see also https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/NPM.20package.20name


grahamegrieve (approved):
ok

PR #498, opened 2022-09-09
When converting to r4 ImplementationGuide, use produceIllegalParameters

dotasek:
See https://github.com/hapifhir/org.hl7.fhir.core/pull/919


grahamegrieve (approved):
ok

PR #508, opened 2022-09-27
new historyPage maintainer, feed conformance fixes, XIG updates

grahamegrieve:


costateixeira:
For the publication process, I'm happy to contribute to testing and
documentation - If I understand correctly, the changes mean the following:

1. The history template expects the current publication folder to contain
the header.template, preamble.template, and postamble.template files
2. When we add the parameter -update-history, the tooling will take the
history template and rebuild the history template page by including those
template files
  2.1. Otherwise it leaves the history page alone?

grahamegrieve:
The templates go at the root, yes. I actually tested this on the HL7 and FHIR web sites. The purpose of -update-history is to update all the history pages without doing anything if you update the templates. see hl7.org.fhir/templates and fhir.org/guides/templates

Issue #108, opened 2020-06-09
Request: use TestScript.scope (or extension in R4) to identify whether a resource has test scripts associated with it


costateixeira:
Goal is to create an optional tab "Testing" for each (conformance) artifact. 
For each given artifact, the Testing tab will be present if there is at least one of the TestScripts in the IG has the artifact's canonical url in the .scope attribute (or an extension with id='scope').
The output would be in resources.json, and could be a flag "testing":"true/false" like history, and/or an array listing the testscripts - "testing":["testscript1", "testscript3"]
The IG template will do the rendering in a similar same way as for history.

Extension draft:
```
<?xml version="1.0" encoding="UTF-8"?>
<StructureDefinition xmlns="http://hl7.org/fhir">
  <id value="replaces"/> 
  <extension url="http://hl7.org/fhir/StructureDefinition/structuredefinition-wg">
    <valueCode value="fhir"/> 
  </extension> 
  <extension url="http://hl7.org/fhir/StructureDefinition/structuredefinition-fmm">
    <valueInteger value="0"/> 
  </extension> 
  <url value="http://hl7.org/fhir/StructureDefinition/scope"/> 
  <version value="4.4.0"/> 
  <name value="scope"/> 
  <status value="draft"/> 
  <date value="2013-12-05"/> 
  <publisher value="Health Level Seven International (FHIR Infrastructure)"/> 
  <contact> 
    <telecom> 
      <system value="url"/> 
      <value value="http://www.hl7.org/Special/committees/fiwg"/> 
    </telecom> 
  </contact> 
  <description value="Indicates the testing scope for a TestScript."/> 
  <fhirVersion value="4.4.0"/> 
  <kind value="complex-type"/> 
  <abstract value="false"/> 
  <context> 
    <type value="element"/> 
    <expression value="TestScript"/> 
  </context> 
  <type value="Extension"/> 
  <baseDefinition value="http://hl7.org/fhir/StructureDefinition/Extension"/> 
  <derivation value="constraint"/> 
   <differential> 
    <element id="Extension">
      <path value="Extension"/> 
      <short value="References a resource that this TestScript is intended to cover"/> 
      <definition value="The resource that is tested by this TestScript."/> 
      <min value="0"/> 
      <max value="*"/> 
    </element> 
    <element id="Extension.extension">
      <path value="Extension.extension"/> 
      <max value="0"/> 
    </element> 
    <element id="Extension.url">
      <path value="Extension.url"/> 
[...]

PR #518, opened 2022-10-11
Add logic to link TestScripts and Profiles

rettema:
Add logic that links TestScript resources to Profiles (StructureDefinitions).

### Why?
The FHIR community has been increasingly interested in provided better support for testing artifacts within IGs. This additional logic is an initial enhancement that allows IGs generated from the HL7 IG template to incorporate and list TestScript resources with their associated FHIR Profiles (StructureDefinitions).

### How?
Minor code additions to two (2) classes that provide the needed links between TestScripts and StructureDefinitions:
* **FetchedResource** 
  * New List\<String\> variable **testProfiles** to hold the canonical url(s) of the FHIR Profile(s) referenced in a TestScript resource; along with corresponding getter and setter methods
  * New HashSet\<FetchedResource\> variable **foundTestScripts** to hold the linked TestScript resource(s) to a StructureDefinition resource; along with corresponding getter and setter methods
* **Publisher**
  * New Set\<FetchedResource\> variable **testscripts** to hold the list of all TestScript resources for the IG
  * **load()** method - new logic in iteration over the IG resources to populate the **testscripts** list; and populate each TestScript's **testProfiles** list
  * **generateSummaryOutputs()** method - new logic to iterate over all **testscripts** and for each linked Profile, update the corresponding StructureDefinition FetchedResource list of **foundTestScripts**
  * **addPageDataRow()** method - add new *testscripts* argument, passed in from profile processing; adds list of linked TestScripts to the Profile page

### DIFF
--- a/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/FetchedResource.java
+++ b/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/FetchedResource.java
@@ -44,7 +44,7 @@ public class FetchedResource {
   private boolean validateAsResource;
   private List<String> statedProfiles = new ArrayList<String>();
   private List<String> foundProfiles = new ArrayList<String>();
-  private List<String> testProfiles = new ArrayList<String>();
+  private List<String> testArtifacts = new ArrayList<String>();
   private boolean snapshotted;
   private String exampleUri;
   private HashSet<FetchedResource> statedExamples = new HashSet<FetchedResource>();
@@ -221,16 +221,19 @@ public class FetchedResource {
     this.path = path;
   }
 
-  public void addTestProfile(String profile) {
-    testProfiles.add(profile);  
+  public void addTestArtifact(String profile) {
+    // Check for duplicate
+    if (!testArtifacts.contains(profile)) {
+      testArtifacts.add(profile);
+    }
   }
 	  
-  public List<String> getTestProfiles() {
-    return testProfiles;
+  public List<String> getTestArtifacts() {
+    return testArtifacts;
   }
 
-  public boolean hasTestProfiles() {
-    return !testProfiles.isEmpty();
+  public boolean hasTestArtifacts() {
+    return !testArtifacts.isEmpty();
   }
 
   public HashSet<FetchedResource> getFoundTestScripts() {
@@ -238,7 +241,14 @@ public class FetchedResource {
   }
 
   public void addFoundTestScript(FetchedResource r) {
-    this.foundTestScripts.add(r);
+    // Check for duplicate
+    if (!foundTestScripts.contains(r)) {
+      this.foundTestScripts.add(r);
+    }
+  }
+
+  public boolean hasFoundTestScripts() {
+    return !foundTestScripts.isEmpty();
   }
 
 }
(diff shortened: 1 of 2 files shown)


costateixeira:
I can't understand the code, but I think that the link from TestScripts should be to any resource. In fact, I think StructureDefinitions would not be among the most common artifacts. The most common ones, I guess, would be 
* CapabilityStatements
* Requirements
* OperationDefinitions
* (other Definitions?)
Can we rename / adjust accordingly? e.g. renaming testProfiles to testArtifacts. and ensuring that this works for any artifact, not Profiles

rettema:
@costateixeira, the TestScript.profile element is the specific link to StructureDefinition that is the basis for this initial enhancement. That being said, yes, the introduction of the TestScript.scope.artifact element in R5 does provide the opportunity to link TestScripts to the other canonical resources. I would prefer addressing this request as a separate enhancement as this will take additional research in conjunction with the HL7 templates for potential display in the IG content.

rettema:
@costateixeira, saw your 2nd comment after sending my reply. Yes, renaming the variable now in anticipation of future use outside of StructureDefinition make sense.

costateixeira:
Actually, I thought we were talking about TestScript.scope - that is the
link to the resource that is being tested, right?

rettema:
TestScript.scope is new for R5 and will represent the full list of linked resources being tested. TestScript.profile has and will remain the specific link to the FHIR Profile (StructureDefinition) used for the validation assert - assert.validateProfileId.

costateixeira:
We had considered preadopting the R5 element as an extension for that. I
think we should have only one entry point - one way of defining how the
TestScript points at what it is supposed to test. If we have more than one,
we need to revisit the resource - I can make a ballot comment for that.

If .profile is indeed be superseded(?)/overlapping with scope, I don't
think we should do this on .profile but instead just do it on the extension
as if it were the .scope element - otherwise I think this will be very
difficult to handle in R5 IGs (which would have to use the wrong element)
or we need to change the code again.
Is it doable in the code to consider the element  .scope preadopted as an
extension?

see https://github.com/HL7/fhir-ig-publisher/issues/108

rettema:
Yes, the logic to parse the TestScript resource for links to conformance resources uses the FHIR org.hl7.fhir.r5.elementmodel.Element representation which allows for a generic (and forgiving) mechanism of traversing the resource structure. So, the code can be written to allow for known variations of where the canonical values are found. This means we can continue to use TestScript.profile along with the new R5 TestScript.scope as well as any potential new extension representing 'scope'.

costateixeira:
So implementers would be able to use either .profile OR .scope in R5? That
is ok for the tooling - but I think the TestScript resource should be
clarified (other discussion).

And for R4, can we use the extension? Happy to contribute to the PR if I
someone can guide me.
I don't think we would ever use the .profile (because it is limiting and
doesn't seem to mean what we need, which is "the conformance artifact that
is covered by this test case".
Thanks

rettema:
> I don't think we would ever use the .profile (because it is limiting and doesn't seem to mean what we need
As I stated above, TestScript.profile is and has been for a very specific use; i.e. assert.validateProfileId. So, it is functionally needed by the Test Engine. I'm not sure we could simply just switch over to TestScript.scope.

In the meantime, let me continue to work on this where we make the TestScript links more generic; i.e. "testProfiles" --> "testArtifacts". I'll keep you updated. Thanks!

grahamegrieve (approved):
ok

PR #524, opened 2022-10-13
Enable -xig execution from Publisher jar

dotasek:
The -xig option in the Publisher jar instantiated XIGGenerator, but didn't actually execute it.


grahamegrieve (approved):
ok

PR #527, opened 2022-10-17
Fix malformed html produced for the ip-statements.xhtml fragment.

rhausam:
Update the SNOMED CT and LOINC text in the getCopyRightStatement() function, and add a strip() statement to the stripPara() function logic so that the ending '\</p\>' suffix is detected and removed.


grahamegrieve (approved):
ok

PR #528, opened 2022-10-17
Catch xig render exceptions

dotasek:
Some packages contain content that can throw exceptions in the XIGRenderer renderResource(...) method.

This catches and reports those exceptions with additional information about the resource that caused them.


grahamegrieve (approved):
ok

PR #533, opened 2022-10-20
Add testscript fragment rendering to StructureDefinitions

rettema:
Add logic to generate TestScript list and table html fragments for Profiles (StructureDefinitions).

### How?
Minor code additions to two (2) classes that provide the needed functionality to generate the TestScript html fragments for StructureDefinitions:
* **StructureDefinitionRenderer** 
  * New method **testscriptList()** to generate the TestScript list html fragment
  * New method **testscriptTable()** to generate the TestScript table html fragment
* **Publisher**
  * **generateOutputsStructureDefinition()** method - new logic modeled after example list and table generation to conditionally call the new **StructureDefinitionRenderer** class methods **testscriptList()** and **testscriptTable()** based on the template config settings **testscript-list** and **testscript-table**.


grahamegrieve (approved):
ok

PR #543, opened 2022-11-02
Change to check if package file does NOT exist - not sure this is the…

costateixeira:
… proper way to negate that, but i tested and seems to work


grahamegrieve (approved):
ok

PR #554, opened 2022-11-18
NPE in PublicationChecker.checkFolder

peturv:
A null-pointer exception occurs if exception does not contain a message. This can happen for example if retrieval was successful, but package-list.json is not parseable as a json file.

Specific scenario is an unpublished IG where the canonical url leads to a web server that has a default redirect to a html page.


grahamegrieve (approved):
ok

PR #574, opened 2022-12-25
native FML support for IG Publisher 

litlfred:
Added support to parse FHIR Mapping Language .fml files.  Features:
* set processing directive --no-fml to true to disable FML
* before scanning for resources, it for .fml files.  for each found, will parse to create a StructureMap resource
* creates temporary directory fml-generated in root of directory in which StructureMap resources (and Library resource containing source .fml) are placed.
*  fml-generated directory is placed into resources directories to make it available for further down the publishing pipeline
*  by default checks input/fml for .fml files, but can be configured with fml-dir configuration parameters

This is a first pass, so feedback appreciated.   Tested running it on https://github.com/litlfred/ig-vcs

 @grahamegrieve @costateixeira


costateixeira:
just a comment from what I can see in the code: The folder name for resources is an IG parameter, and there is already a "maps" folder that gets added automatically in the base template: https://github.com/HL7/ig-template-base/blob/master/scripts/onLoad.xslt#L163
should we default to that folder and just look for files with extension .fml in there? 
Or do we want to create a folder exclusively for fmls?

grahamegrieve:
it should be the maps folder, yes, that's the intention. in fact, I thought it already worked that way

litlfred:
There is only very limited support for maps (mostly about generating maps form content in StructureDefinition extensions) as is here:
    https://github.com/litlfred/fhir-ig-publisher/blob/master/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java#L6120

There is nothing that processes FHIR Mapping Language files that I could find anywhere.

Per @costateixeira  and @grahamegrieve  comments - I can adjust to use whichever (default) directories as desired such as input/maps.  However, that seems to be simply a directory for where it should look for source files for processing concept maps and spreadsheets.  What is not there is where produced StructureMaps should be saved to.   Specifically, this pull request does the following in the spirit of getting more and more computable assets:

- looks for any fml files
- processes the fml to produce StructureMap resources
- creates a Library resource that holds the source fml
- creates linkages between the SM and Library reosurces


Also,  see the comments at the end of the generateOutputsStructureMap() for some related to-dos.
https://github.com/litlfred/fhir-ig-publisher/blob/master/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java#L10217
as well as the comment here:  https://github.com/litlfred/fhir-ig-publisher/blob/master/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java#L5135


I am working on another feature branch to generate UML files to illustrate StructureMaps.   This would work on either of the SMs generated from FML or ones that already exist.

litlfred:
@costateixeira  on other reason I was hesitant in making use of the path-resource directory is that, in analogue to how sushi is handled,  I wanted the fml's to be processed into structure maps before the resource directories are processed. 

See around here:
   https://github.com/litlfred/fhir-ig-publisher/blob/master/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java#L4176
for the overall sequencing.

costateixeira:
In my previous comment was assuming the fml files were "normal" files, so we'd use the folder "mappings" as a resource folder, and still ask the publisher to look into all the path-resource folders for .fml files

But if fmls are handled differently from the other resources, using a special parameter about "where to find .fml files" seems good. This is a detail that I'd leave to @grahamegrieve to resolve so that this can be unblocked and tested

lmckenzi:
FML files should be processed the same as JSON or XML files.

litlfred:
@lmckenzi  not exactly sure what you mean as there are few place impacted and would appreciate further guidance.

1)  I had already modified the publisher.parseContent() function to handle fml similar to json and xml so that's there.  Iam using mime-type "text/fhir-mapping" which is what @oliveregger  is using on matchbox, but don't know if that's canonical,

2) i had already modified there is a simplefetcher.scan() function to handle fml similar to json and xml.  as an aside, probably some of the existing logic for xml, json, spreadsheets, around which files are ignored could be cleaned up there but i don't want to break anything - @grahamegrieve  @jamesagnew  i was slightly confused by the checks for files extensions not being in a list rather than checking that file extensions are in a list before try parsing.  in any case, I put an explicit check for .fml filename extension for handling.

3) there is a publisher.loadFML() function called by the publisher.load() function, the reasoning for putting it before the json and xml are processed was that fml seemed more analogous to pre-processing fsh files to generate resources and that we would want these available (immediately) before we call loadResources().  Noting that in the case of fml we generate two resources per fml file  ( StructureMap and Library containing the source fml ) there are a few options here:
  a) sushi is called by the initialize() so, if we wanted to follow the sushi analog closely we would probably want to move the loadFML() logic to be called by initialize().   
  b) have loadFML() incorporated into the loadResource() logic.   this would mean to modifiy simplefetcher.scan() so that when it sees an fml file in produces both the structuremap and the library.   currently it only creates the structuremap.  however it seems that scan() associates one resource per file so don't know what this could potentially mess up.
  c) leave the loadFML() as is

lmckenzi:
We think the issue might be one of file extension - .map instead of .fml.  @costateixeira will work to help test.  If so, all we need to do is improve documentation.

litlfred:
If you mean simply changing my_file.fml to my_,file.map and making sure
everything is under input/map that it would work, then I just tested and it
doesn't.

Cheers,
-carl

costateixeira:
confirming that putting a .map file in input/resources does NOT create the
structuremaps

grahamegrieve:
Why is there a library resource needed?

grahamegrieve:
Well, we do want comments to round trip. They definitely should. I’m going to do an implementation of what it should do over the next week and then we can reconsider

litlfred:
Issue will be that there are no places to put the comments in the
structuremap (except in the description which doesn't preserve location of
the comment s).  I guess an extension that could be applied to every node
in the SM, but that seems less than ideal and perhaps runs into issues of
where the comments go in the FML relative to the various tokens.

grahamegrieve:
I think this is all working now from other changes - please test and if there's still issues, discuss before making a PR about it

PR #576, opened 2023-01-06
Test XSL Transformation

dotasek:
This PR refactors XSL Transformation into a testable class and verifies that the XSL transformation module included in the JRE is one that doesn't produce extra namespace prefixes.

### DIFF
--- a/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java
+++ b/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java
@@ -65,19 +65,12 @@ import java.util.zip.ZipInputStream;
 import javax.swing.UIManager;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
-import javax.xml.transform.ErrorListener;
-import javax.xml.transform.Transformer;
-import javax.xml.transform.TransformerException;
-import javax.xml.transform.TransformerFactory;
-import javax.xml.transform.stream.StreamResult;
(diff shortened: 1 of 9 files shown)


grahamegrieve (approved):
ok. I mean, this is a good as a test, but it seems to me that it doesn't quite hit the spot, since whether it falis or not isn't due to our code, but due to the platform provider? Would it be good to run this test at the start of the IG publisher run?

dotasek:
The change in core SHOULD ensure that there is no platform dependency. It explicitly uses Saxon, as it did before, when we got Saxon as a side-effect from HAPI. That will override the platform.

Maybe end users could go through the effort of configuring another TransformerFactory implementation when running IG publisher to override this, but I think it's an unlikely thing, and that in that case the test would interfere with their intent.

grahamegrieve:
Well, all those things are true but what does this test actually test then? Does it fail if Saxon isn’t included?Sent from my iPhoneOn 7 Jan 2023, at 10:32 am, dotasek ***@***.***> wrote:﻿
The change in core SHOULD ensure that there is no platform dependency. It explicitly uses Saxon, as it did before, when we got Saxon as a side-effect from HAPI. That will override the platform.
Maybe end users could go through the effort of configuring another TransformerFactory implementation when running IG publisher to override this, but I think it's an unlikely thing, and that in that case the test would interfere with their intent.

dotasek:
@grahamegrieve If saxon isn't included, this test MIGHT fail. It definitely fails on my machine. I imagine, as your computer didn't exhibit the same behaviour running the ig-publisher on the unicom IG, that there are some JRE distributions that could still pass.

grahamegrieve:
ok

PR #578, opened 2023-01-09
Restore surefire tests

dotasek:
This restores surefire tests for the pipelines, which were disabled.

AllGuidesTests is now disabled unless the test IGs directory exists


dotasek:
Merging after approval from @grahamegrieve on Zulip

PR #587, opened 2023-01-12
Add GraalVM js engine

dotasek:
Since Java 15, there is no longer a default JavaScript engine within the JVM.

The publisher uses Ant build files which are processed via org.apache.ant, which requires a JavaScript engine.

This adds GraalVM as the default engine for this purpose via dependencies, and should allow Java 11 and 17 to run JavaScript script within Ant build files.

**This is presently only working via direct execution in IntelliJ. Somehow, the process of packaging Graal into the CLI jar is not working correctly.**

### DIFF
--- a/org.hl7.fhir.publisher.core/pom.xml
+++ b/org.hl7.fhir.publisher.core/pom.xml
@@ -12,6 +12,10 @@
 
     <artifactId>org.hl7.fhir.publisher.core</artifactId>
 
+    <properties>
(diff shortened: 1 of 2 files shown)


grahamegrieve (commented):
works in eclipse

dotasek:
Superseded by this PR: https://github.com/HL7/fhir-ig-publisher/pull/613

PR #626, opened 2023-03-03
add fml support for StructureMap in ig

oliveregger:
this PR extends the IG support that
- fml files are converted to StructureMap resources and included in the IG package
- fml are converted with R5 (not only R4 and R4B)

goal is to build an ig (another PR for the packages/inteversion/r5) will follow

i'm not very familiar with the inner workings of ig publisher, had to add the map parsing at 3 different places, so a review, adapation would be appreciated


oliveregger:
i tried to create an Implementation Guide for https://github.com/FHIR/packages/tree/master/interversion/r5 that I can use the maps as FHIR R5 Implemenation Guide to transform from R4, R4B to R5 and back. I came up with the following setup https://github.com/FHIR/packages/pull/13 but I had to adjust the IG Publisher that it was working. If there is an easier way, I'm happy, I just need and IG with all the StructureMaps inside :-)

grahamegrieve (approved):
ok

PR #638, opened 2023-03-14
Fix HTMLInspector paths that fail in Windows

dotasek:
Utilities.path constructs java Path instances to ensure that Paths aren't doing anything nefarious, but the two instances here construct 'kinda' paths with filesystem paths that have anchors ('C:/myig/somefile.html#extension:somethingelse') which can contain illegal characters in the Windows OS. 

Utilities.path should have a way to deal with this, but until then, this works around it.


dotasek:
Approved by @grahamegrieve in weekly meet.

PR #650, opened 2023-03-20
Fix FSH ini fetching

dotasek:
FSH ini fetching would fail if the IG directory was in a root directory: `C:\myIG`

If it failed to be in the initial directory, it would attempt to build a path with the parent, which would result in a Path exception. ~~This catches and logs that exception, and continues to execute (I think it's not a great use case to have fsh in the root directory).~~ This fix doesn't attempt to build a path with the parent.


dotasek:
Merging pending conversation with @grahamegrieve on Zulip.

PR #669, opened 2023-04-06
Prevent checks for URL templates as directories

dotasek:
When a template is a URL, it will resolve as a path under Linux/Max filesystems, but will break in Windows.

This skips attempting to check URLs as possible local directories when loading a template.


dotasek:
Merging with approval from @grahamegrieve on zulip

PR #671, opened 2023-04-13
Set FhirSettings via param in Publisher + use FhirSettings everywhere

dotasek:
This uses the draft FhirSettings instead of ini settings.

See: https://github.com/hapifhir/org.hl7.fhir.core/pull/1203


dotasek:
Superseded by https://github.com/HL7/fhir-ig-publisher/pull/677

PR #682, opened 2023-05-04
Suppress mappings that are totally empty

stueynz:
If we suppress mappings that are all empty - then the `-maps-diff` HTML fragment will contain only the mappings that you've added in your differential. 

Make use of custom template to use `-maps-diff` rather than `-map` HTML fragment gets a much more useful Mappings tab in the StructureDefinition.

While I'm here, I made the `comment` column a markdown field.


grahamegrieve (approved):
ok

Issue #618, opened 2023-02-21
Add TestPlan to IG (as we have for TestPlan)


costateixeira:
When an artifact has a TestPlan pointing to it via .scope, the same Testing page should open as for the TestScript. 
The content of the tab will be a rendering of the test plan, and then the testscripts.

lmckenzi:
Agree in principle.  Will formally approve once there's a mock-up of what it looks like.

PR #694, opened 2023-05-23
#618 Additional update to include TestPlan in prior FHIR versioned IGs

rettema:
Additional update to include TestPlan in prior FHIR versioned IGs

### Why?
Related to: [#618](https://github.com/HL7/fhir-ig-publisher/issues/618)


lmckenzi (approved):
Ok

PR #716, opened 2023-06-26
Support private (non-HL7) publication of templates

ElliotSilver:
This PR removes the requirement that templates are published to the fhir.org site.

It consists of 3 changes.
- Previously, all templates were published by HL7 from a single source tree, the publisher could enforce that templates which depended on other templates would only depend on the most recent version. A template could use the "current" version of another template, and this gets converted to the most recent specific version on publication; and reverted back after.
  Because templates published to sites other than HL7 won't have all those parent templates in the publication tree, that check isn't possible for private publication. In this PR, the check is skipped if an environment variable is set, and specific versions of parent templates must be listed.
- The template index page makes assumptions about the site address and structure. This is made more dynamic by reading that information from package-list.json.
- The content of the index page also includes HL7-specific text. An environment variable specifying an a file containing alternate content is now supported.

Additionally, a couple of spelling errors were corrected.

### DIFF
--- a/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/utils/TemplateReleaser.java
+++ b/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/utils/TemplateReleaser.java
@@ -123,6 +123,8 @@ public class TemplateReleaser {
 
   private File xml;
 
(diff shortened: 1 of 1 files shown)


lmckenzi:
Why package-list.json?  Isn't that replaced by publication-request.json?

lmckenzi:
@ElliotSilver will review for next week

PR #727, opened 2023-07-26
Fix invalid filename path check in DICOM

dotasek:
SpecMapManager is running a special case check for files in DICOM, but asks for some filenames invalid in Windows.


grahamegrieve (approved):
ok

PR #728, opened 2023-07-27
Compile fix

grahamegrieve:


dotasek:
Mergine after discussion with @grahamegrieve on Zulip

PR #735, opened 2023-08-11
Add options to IG Publisher UI

dotasek:
Adds the following options to the IG Publisher UI:

* debug
* no-sushi
* no-validate
* no-narrative

In addition, this fixes an underlying issue with no-validate and no-narrative params, where there was a typo in in the param (asking for a 'non-validate' and 'non-narrative' param).


grahamegrieve (approved):
looks good

PR #736, opened 2023-08-11
Toc fixes

lmckenzi:
NOTE: Requires Grahame to push a change to the pubpack.


grahamegrieve (commented):
why the changes to publisher.java?

lmckenzi:
The only thing changed is publisher.java - that's where the ToC is defined.

Presume you mean `else if (defaultBusinessVersion != null && bc.getVersion()==null)`?
That's because the prior code was failing when bc.getVersion() was null and throwing an NPI.

grahamegrieve:
The correct fix is else if (defaultBusinessVersion != null && !bc.getVersionElement().isEmpty())

lmckenzi:
Fixed

grahamegrieve (approved):
ok

PR #755, opened 2023-09-10
SQL implementation

grahamegrieve:


org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/renderers/DBBuilder.java
```diff
@@ -0,0 +1,535 @@
...
+        "PRIMARY KEY (Key))\r\n");
+  }
+
+
+  public String processSQL(String sql) {
+    if (con == null) {
+      return "<span style=\"color: maroon\">Error processing SQL: SQL is not set up properly</span>";
+    }
+
+    sql = sql.trim();
```
dotasek, line 413: Null sql here could blow up. I'd suggest doing something like this in the catch block:

```java
 if (sql == null) {
        throw new IllegalArgumentException("Param sql cannot be null.");
      }
     
      sql = sql.trim();
```

dotasek (changes requested):

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/renderers/DBBuilder.java
grahamegrieve: ok.

dotasek: I moved that into the try block. Feel free to revert that commit if it's improper, but I think it's safer to do it that way.

grahamegrieve: no that's ok

PR #760, opened 2023-09-15
Fixed issue where package ids that are hl7.cda and hl7.v2 were spitti…

lmckenzi:
…ng out error messages

Also fixed https://jira.hl7.org/browse/HSCR-477 - needed to spit out all profiles, not only the first.


grahamegrieve (commented):
ok

PR #796, opened 2023-10-29
Updated CQL translator version to 3.3.2

brynrhodes:
This PR updates the CqlSubsystem to use the latest release of the CQL-to-ELM translator, version 3.3.2. This update was tested with the sample-ig and sample-content-ig implementation guides and confirmed that it functions correctly. This update addresses issues with resolving references to FHIRHelpers and FHIR ModelInfo, as well as some improvements to the output of the Library refresh. Note that this update does not make use of the DataRequirementsProcessor since that would require introducing a dependency on the elm-fhir package from the translator.

### DIFF
--- a/org.hl7.fhir.publisher.core/pom.xml
+++ b/org.hl7.fhir.publisher.core/pom.xml
@@ -15,7 +15,7 @@
     <properties>
         <graalvm.version>22.3.3</graalvm.version>
         <jackson_version>2.14.0</jackson_version>
-        <info_cqframework_version>1.5.12</info_cqframework_version>
+        <info_cqframework_version>3.3.2</info_cqframework_version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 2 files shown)


dotasek:
As a test, I ran this against the `br-translator-3.3.2-updates` branch of `sample-ig`

At very least, it does appear to complete, and it does reduce the number of errors reported by the publisher:

old
```
Errors: 12, Warnings: 10, Info: 2, Broken Links: 0 (00:00.202 / 01:50.332, 747Mb)
```

3.3.2
```
Errors: 10, Warnings: 13, Info: 2, Broken Links: 0 (00:00.166 / 02:24.116, 786Mb)
```

I should probably have saved the qa.html output to compare. I will do that next.

It also doesn't appear to change the console outputs of the other 4 IGs I routinely test against. 

Is there some specific way to verify that this code actually does the right thing @brynrhodes? Areas in the IG output, etc?

grahamegrieve (approved):
ok

brynrhodes:
@dotasek , I don't have any programmatic tests at this level no, I verify the functionality manually with the sample-ig and sample-content-ig IGs, as well as several of the other project IGs like Opioids and eCQM-Content.

PR #802, opened 2023-11-02
Plot builder pipeline

dotasek:
Adds the following script to a pipeline that runs parallel to the master build pipeline:

https://github.com/djvreeman/hl7-standards-dev-data/blob/main/scripts/plot-ig-builder-auto.py

This updates the following file in the GitHub repository: https://github.com/HL7/fhir-ig-publisher/blob/master/plot-ig-builder-auto.png


grahamegrieve (approved):
looks good to me as far as I can follow, but what happens if the .py fails? the pipeline fails?

dotasek:
@grahamegrieve It's a separate pipeline that should get triggered the same way the master build and deploy pipeline runs. If it fails, it will be an 'X' on the build, but the SNAPSHOT should still get deployed.

PR #808, opened 2023-11-08
When autoloading document Bundles, read title from Composition

adamzkover:
Support autoload of some Bundles. If the Bundle is a document, use the title from the Composition as the resource title in the IG.
This does not make a difference for any other resource type and is a better alternative to not having a title at all.


grahamegrieve (changes requested):
most bundles won't be resources, so you should use the r.getElement() not the r.getResource()

adamzkover:
> most bundles won't be resources, so you should use the r.getElement() not the r.getResource()

The change is now rewritten to use r.getElement().

grahamegrieve (approved):
ok

PR #810, opened 2023-11-08
Update plot with dynamic width and better colors

### LINKED ISSUE #809: Use unique colors and dynamically set image width based on number of versions in the data
Presently, the plot re-uses some colors for different guides. It'd be best to have a unique color per guide.

Also, the image will not look very good once the number of versions grows large. Let's dynamically adjust the plot based on the number of versions per guide.

djvreeman:
Resolves #809 

Modifications to plot-ig-builder-auto.py:
- Adds new color map (up to 40 colors with room for 60)
- Sorts the figure legend by performance time so that its easier to match the guide with their times
- Adds dynamic width generation and some visual tweaks to adjust the plot as we get more versions. Tested with [fake data](https://github.com/djvreeman/hl7-standards-dev-data/blob/main/data/extended-statistics-fake.json) up to about 50 versions.


djvreeman:
FYI, this is what the plot looks like with fake data times for a bunch of releases:
plot-ig-builder-auto-fake

grahamegrieve (approved):
ok

PR #832, opened 2023-12-20
FIX: The eslint text file does not contain all the error's, warnings and informationals.

hvamstel:
The generation of the qa-eslintcompact.txt does not contain all the errors, warnings and informationals that are generated.
The ones that are called "linkerrors" were not present in the output.

We use this file to fill the problem window of visual code so that the user can just click on an error and is presented with the correct file and location automatically.

This pr completes this eslintcompact with that addtional information so that it is complete and contentwise identical to the other qa outputs. Now the total number of errors, warnings and informationals is correct with the summary given.


grahamegrieve (approved):
ok

PR #852, opened 2024-01-24
The button tag is not auto-closing

### LINKED ISSUE #848: Detailed Descriptions: content following the 'copy URL' button is badly rendered
In the 'Detailed Descriptions' tab of resources, when there is content following a 'copy URL' button, it is badly rendered:

build fhir org_ig_hl7ch_ch-core_StructureDefinition-ch-core-composition-definitions html
build fhir org_ig_hl7ch_ch-core_StructureDefinition-ch-core-patient-definitions html

E.g. in https://build.fhir.org/ig/hl7ch/ch-core/StructureDefinition-ch-core-patient-definitions.html (built with v1.5.14)
https://github.com/hl7ch/ch-core

qligier:
As per https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#technical_summary:

> Tag omission:	None, both the starting and ending tag are mandatory.

Does not fix #848 (yet), because the HTML is parsed and re-rendered wrongly by `XhtmlComposer`. Another patch in core is needed: https://github.com/hapifhir/org.hl7.fhir.core/pull/1553


grahamegrieve (approved):
ok

PR #889, opened 2024-04-24
add ch.fhir.ig.template as trusted

oliveregger:
see discussion in https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Trusted.20Template.20Process for additional information.

email informing committer/grahame about process has be done


grahamegrieve (commented):
ok per email with committers

PR #891, opened 2024-05-05
Add classes to profile header content

jmandel:
I have scenarios where I'd like to style content differently, and selecting `<p>` without classes won't work. Is something like this viable?

---

Also, there may be a simpler way for me to accomplish what I want, which is to do something like

    {% include StructureDefinition-cgm-data-submission-bundle-header.xhtml %}

...in my markdown pages without pulling in anything except the markdown description of my profile.

I still think assigning classes in generated HTML is a good idea though.

### DIFF
--- a/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/renderers/StructureDefinitionRenderer.java
+++ b/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/renderers/StructureDefinitionRenderer.java
@@ -1300,9 +1300,9 @@ public class StructureDefinitionRenderer extends CanonicalRenderer {
       b.append("<p class=\"profile-derivation\">\r\n");
       StructureDefinition sdb = context.fetchResource(StructureDefinition.class, sd.getBaseDefinition());
       if (sdb != null)
-        b.append(/\*#!\*/"This profile builds on" + " <a href=\\"" + Utilities.escapeXml(sdb.getWebPath()) + "\\">" + gen.getTranslated(sdb.getNameElement()) + "</a>.");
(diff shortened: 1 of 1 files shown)


org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/renderers/StructureDefinitionRenderer.java
```diff
@@ -1289,23 +1289,23 @@ private List<ElementDefinitionMappingComponent> getMap(ElementDefinition e, Stri
...
   public String header() throws Exception {
     StringBuilder b = new StringBuilder();
-    b.append("<p>\r\n");
+    b.append("<p class=\"profile-url-label\">\r\n");
     b.append("The official URL for this profile is:" + "\r\n");
     b.append("</p>\r\n");
-    b.append("<pre>" + sd.getUrl() + "</pre>\r\n");
-    b.append("<p>\r\n");
+    b.append("<pre class=\"profile-url\">" + sd.getUrl() + "</pre>\r\n");
+    b.append("<div class=\"profile-description\">\r\n");
```
jmandel, line 1296: changed `<p>` to `<div>` because the markdown can contain `<p>` which I don't think should be surrounded by outer `<p>`

grahamegrieve (commented):
what's with changing the <p> to a <div>?

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/renderers/StructureDefinitionRenderer.java
grahamegrieve: if it doesn't contain a paragraph, does this change the way it displays?

jmandel: It depends how the stylesheet handling of `p` but generally paragraph don't have any margin etc.

PR #904, opened 2024-05-24
Select rendering context locale from IG

dotasek:
Selects the rendering context locale from the source IG instead of defaulting to the JVM locale.


grahamegrieve (approved):
ok

Issue #910, opened 2024-06-18
Double entries in IG when providing FHIR resources as binaries


oliveregger:
For testing we would like to provide non conformant FHIR resources within an IG. We wan't to add them as Binary resource, so that the validator / IG Publisher does not create errors or warnings for it and that it can be also processed. 

This works fine and is no problem if we provide the FHIR resource base64 encoded directly in the binary.

However if use the nice FHIR IG Publisher shortcut and use the automatic Binary Conversion (like for logical models) we get two entries, one for the binary (which what we wanted) but also one for the resource itself, see for example:

https://build.fhir.org/ig/ahdis/ch-ig/branches/oe_binary/qa.html

we have there now two entries:

<img width="828" alt="image" src="https://github.com/HL7/fhir-ig-publisher/assets/2414516/c88a0b6d-53ad-4026-bae4-00bb2de635bf">

but we would only like to have one, when we would create it manually.

source of branch/repo reproducing it is here: https://github.com/ahdis/ch-ig/tree/oe_binary

zulip background discussion is here: https://chat.fhir.org/#narrow/stream/274423-TestScript-Resource/topic/How.20to.20test.20invalid.20data

will provide a PR.

lmckenzi:
This appears to now be fixed.  @oliveregger can you confirm and, if so, close?

PR #911, opened 2024-06-18
do not add Binary resources which are not really Binary #910

oliveregger:
see #910


grahamegrieve:
I cannot be happy with this PR as it is. My main concern is not what it does but what it undoes. What other IGs are going to be broken by this change?

oliveregger:
In addition to verify it with the provided CH-IG where the duplication happened, i tested it with the IG where the binary approach with logical model examples are defined via the binary mechanism, and this works still as expected:

image

compare to: https://build.fhir.org/ig/ahdis/ch-alis/branches/__default/qa.html

do you know of any other IG I should/could test?

PR #928, opened 2024-07-23
Tweaks to the codepaths related to the cross version IG pre-processor

brianpos:
Specifically:
* remove the hard-coded path to `/Users/grahamegrieve/work/fhir-cross-version` and pass the path via parameters
* ensure the `temp/xver-qa` path exists before trying to read/write to it


grahamegrieve:
I think we want to get rid of all those saving files that have hardcoded file paths

brianpos:
> I think we want to get rid of all those saving files that have hardcoded file paths

Agree with that, but do need the other change here with the checking that the `temp/xver-qa` path exists...
(that's what's killing the package CI build at the moment)

PR #932, opened 2024-07-26
Delete unneeded workflow

costateixeira:
With the new project and [issues board](https://github.com/orgs/HL7/projects/6/views/1) and the deprecation of the classic projects and respective boards, the  action to automatically add issues is no longer needed nor working.


grahamegrieve (approved):
ok

PR #938, opened 2024-08-07
Fix issues with UTF-8 2-byte characters

dotasek:
The previous implementation of MySushiHandler used a StringBuilder, which was trying to append the first byte value of a two-byte UTF-8 character, resulting in a silent failure.

Fix uses a ByteArrayOutputStream and also treats invalid characters b replacing them with '?'.


grahamegrieve (approved):
ok

PR #991, opened 2024-11-20
Fix binding table order

lmckenzi:
Replace https://github.com/HL7/fhir-ig-publisher/pull/986 (made it stand-alone)


grahamegrieve (approved):
ok

Issue #888, opened 2024-04-24
Key elements tab missing some elements in differential


RichardTON:
The key elements table should include elements that are in the profile's differential.

For example [AU Base Patient](https://build.fhir.org/ig/hl7au/au-fhir-base/StructureDefinition-au-patient.html) includes constraints on [Patient.contact](http://patient.contact/). [Patient.contact](http://patient.contact/) does not appear in the key elements table.

See Zulip https://chat.fhir.org/#narrow/stream/179252-IG-creation/topic/Make.20KeyElements.20the.20default.20tab/near/434905716

lmckenzi:
We will update key elements to show an element if the type constraints or bindings have changed, even if those elements aren't MustSupport.

Side note: Specifying types of "Address, AUAddress" in the differential doesn't actually accomplish anything from a validation perspective.  (Though it does highlight the availability of those other profiles if implementers happen to care.)

RichardTON:
> Side note: Specifying types of "Address, AUAddress" in the differential doesn't actually accomplish anything from a validation perspective. (Though it does highlight the availability of those other profiles if implementers happen to care.)
This is recognised. The intent is to advertise useful profiles and simplify creating constraints using them in derived profiles.

PR #996, opened 2024-11-21
Key elements enhancements

lmckenzi:
Addresses
https://github.com/HL7/fhir-ig-publisher/issues/888

Also ensures that issues with SQL or fragment generation always put something in the log


grahamegrieve (approved):
ok

PR #1021, opened 2025-01-17
Hacky POC of fixing FHIR package transitive dependency problem

bkaney:
This was a problem discovered when trying to refer to dependant CQL libraries in external FHIR packages. The npm list was not including transitive dependencies.


grahamegrieve:
wel, yes, hacks - it didn't actually fix the problem. Fixed elsewhere

bkaney:
Thanks, I just tested origin/master with my test cases and can confirm it is now fixed!

PR #1022, opened 2025-01-23
Disallow incorrectly formatted Git URLs

dotasek:


grahamegrieve:
why don't we just strip the user information, instead of returning null? The repoSource has legitimate use in the pipeline, and why should the presence of user info from git blow the pipeline?

dotasek:
Addressed comments, @grahamegrieve let me know if this is good to merge.

grahamegrieve (approved):
looks good now, though the javadoc needs updating

PR #1020, opened 2025-01-13
Updated publisher to use the latest CQL translator, 3.20.0

brynrhodes:
Tested with sample-ig, cql-ig, and ecqm-content-qicore-2025

PR #1042, opened 2025-02-14
Fix for CQL Framework Dependencies

dotasek:
This PR removes the unpack-dependencies execution and uses dependencyManagement to solve the dependency issues caused by conflicting JAXB libraries. 

This should supersede https://github.com/HL7/fhir-ig-publisher/pull/1020

It is presently tested with sample-ig, but reviewers should also verify that cql-ig and ecqm-content-qicore-2025 IGs still publish correctly, as well as the usual IG testing set.


dotasek:
Also confirmed to work with the templating step in UTG, which has been known to have issues when the build changes due to Graal dependencies.

brynrhodes (approved):
Tested and confirmed working as expected with sample-ig, sample-content-ig, cql-ig, and ecqm-content-qicore-2025

grahamegrieve (approved):
looks ok to me

PR #1084, opened 2025-04-24
Delta Workflow

dotasek:
Adds a manually triggered workflow that compares the QA outputs of multiple common IGs.


grahamegrieve (approved):
looks ok to me

PR #1087, opened 2025-04-29
Use relative filenames for comparison QA

dotasek:


dotasek:
No longer needed. This was originally intended to allow comparisons between QA results that were executed locally and on the pipelines. However, the new workflow runs all it's diffs on workflows, making this unnecessary.

https://github.com/HL7/fhir-ig-publisher/pull/1084

PR #1090, opened 2025-05-06
2025 05 gg qa speed fix

grahamegrieve:


dotasek (changes requested):
The speed changes look reasonable.

However, there are a few things that need addressing. 

1. The big one is that JavaFX is licensed under GPL 2.0

https://mvnrepository.com/artifact/org.openjfx/javafx-base/24

We are under an Apache License 2.0, which is incompatible: https://www.apache.org/licenses/GPL-compatibility.html

2. The added maven-shade-plugin section in the root pom.xml breaks the existing shading plugin operation when the project is built. Removing it seems to fix the build, and the '-gui' option runs fine without it.

3. There's a missing constant: RenderingI18nContext.SDR_EXTENSION_SUMMARY, which shows up in the IG output as a missing message. Maybe something that wasn't committed, or was accidentally deleted from core.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/ui/IGPublisherUI.java
```diff
@@ -0,0 +1,775 @@
...
+            
+            try {
+                // Redirect System.out and System.err to our custom stream
+                System.setOut(customPrintStream);
+                System.setErr(customPrintStream);
+                
+                // Call the Publisher directly
+                Publisher.runDirectly(
+                    folderPath,
+                    termServer.isEmpty() ? null : termServer,
```
dotasek, line 601: Setting the terminology server to null here causes this to run without a terminology server, instead of the default server.

I think it should be this:

```java
termServer.isEmpty() ? FhirSettings.getTxFhirProduction() : termServer,
```

org.hl7.fhir.publisher.core/src/main/resources/styles.css
dotasek, line 1: I think this is leftover from JavaFX, and should be removed. 

When I remove it, I can still run the UI.

dotasek (changes requested):

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/ui/IGPublisherUI.java
grahamegrieve: removed

dotasek (approved):

PR #1137, opened 2025-07-24
Run XIG pipeline in one step

dotasek:
This runs all of the XIG pipeline using a single step with a longer timeout.


dotasek:

PR #1146, opened 2025-08-08
Bump org.eclipse.persistence.moxy and jaxb-runtime

dotasek:


brynrhodes (approved):
Verified with hl7.fhir.uv.cql and hl7.fhir.us.cql

PR #1147, opened 2025-08-11
Publisher refactor cleanup

dotasek:
Some renaming and refactoring:

Renaming f to publisherFields
Moving getters and setters into publisherFields
Moving inner classes up


grahamegrieve (approved):
ok

PR #1168, opened 2025-10-01
2025 10 gg minor fixes

grahamegrieve:


org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/renderers/StructureDefinitionRenderer.java
```diff
@@ -1035,33 +1035,37 @@ public void txItem(Map<String, ElementDefinition> txmap, XhtmlNode tbl, String p
...
 
   private String presentVersion(PackageInformation sourcePackage) {
-    switch (sourcePackage.getCanonical()) {
-      case "http://fhir.org/packages/fhir.dicom":
-        String[] vp = sourcePackage.getVersion().split("\\.");
-        return vp[0]+String.valueOf((char)('a' + (vp[1].charAt(0) - '1')));
-      default:
-        return VersionUtilities.getMajMin(sourcePackage.getVersion());
+    if (sourcePackage.getCanonical() != null) {
+      switch (sourcePackage.getCanonical()) {
```
dotasek, line 1039: Could this be a simpler if statement instead of a switch here?

dotasek (changes requested):
One minor nitpick.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/renderers/StructureDefinitionRenderer.java
grahamegrieve: it could be, but it's likely to grow and it matches the previous function structurally.

PR #1173, opened 2025-10-05
2025 10 gg expansions

grahamegrieve:


org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/AIProcessor.java
```diff
@@ -0,0 +1,429 @@
...
+      if ("div".equals(c.getName()) && hasTextNotInDiv(c, s)) {
+        toDelete.add(c);
+      } else {
+        stripDiv(c, s);
+      }
+    }
+    x.getChildNodes().removeAll(toDelete);
+  }
+
+  private boolean hasTextNotInDiv(XhtmlNode c, String s) {
```
dotasek, line 330: The logic here seems to skip the top level div if I'm not mistaken. Should there not be a check for `c` as well?

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/AIProcessor.java
dotasek, line 1: SonarQube flagged some code in OldIGAIGenerator as duplicating chunks of this. Can OldIGAIGenerator be removed?

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/AIProcessor.java
```diff
@@ -0,0 +1,429 @@
...
+      }
+      FileUtilities.stringToFileIfDifferent(md, Utilities.path(dir, dest));
+      zip.addFileSource(dest, md, true);
+      return md.length();
+    } catch (Exception e) {
+      throw new FHIRException("Unable to process " + p.f.getAbsolutePath(), e);
+    }
+  }
+
+  private boolean isText(String contentType) {
```
dotasek, line 310: This could just be:

```java
private boolean isText(String contentType) {
      return contentType.startsWith("text/");
  }
```

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/AIProcessor.java
```diff
@@ -0,0 +1,429 @@
...
+    for (Page p : pages) {
+      if (!p.f.getName().equals("index.html") && p.kind == PageKind.PAGE) {
+        newSize += produceMDForPage(llms, p, zip, npm);
+      }
+    }
+    llms.append("\r\n## Resources\r\n");
+    resources.sort(new LoadedResourceSorter());
+    String t = null;
+    for (LoadedResource r : resources) {
+      Page p = getPageForResource(r, pages);
```
dotasek, line 75: I get a warning saying p could be null at this point. Should there be an assert or check here?

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/AIProcessor.java
```diff
@@ -0,0 +1,429 @@
...
+
+  private File file(File dir, String s, String ext) throws IOException {
+    return ManagedFileAccess.file(Utilities.path(dir, FileUtilities.changeFileExt(s, ext)));
+  }
+
+  private File file(File dir, String s) throws IOException {
+    return ManagedFileAccess.file(Utilities.path(dir, s));
+  }
+
+  public void processFolderOld(File dir, NpmPackage npm) throws IOException {
```
dotasek, line 103: Add some comments for what 'process' means, or change the name to be clearer.

The following would be a little too verbose, but at least a comment to indicate intent on these public methods would give some hints for future development.

generateMarkdownForPagesAndIndexInLLMS...

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/AIProcessor.java
grahamegrieve: maybe, and it is duplicative, but not yet - there's some reason I might run it again, and it's just not worth trying to merge the code

PR #1178, opened 2025-10-25
GitHub files should not block template

### LINKED ISSUE #1172: .github files incorrectly detected as untrusted file extensions
When the repo has any content in .github (e.g. .github/CODEOWNERS) the ci build is blocked

Image

I can expect some cases where .github content is expected (workflows, issue templates, etc) and the CI build should not block because of those

costateixeira:
Fixes #1172


grahamegrieve (approved):
ok

PR #1205, opened 2025-12-05
Fix support for SSH-style Git URLs (fixes #1204)

### LINKED ISSUE #1204: Does not support git url like git@github.com:org/repo.git
For security reason, all my git are pulled with url such as git@github.com:org/repo.git (default use SSH key)

but the publisher failes here

https://github.com/HL7/fhir-ig-publisher/blob/cccfbc02fb03141c81e4b417a467d39604128312/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/GitUtilities.java#L62

`Warning @ Git URL is not a valid URl. Source: git remote origin ur`

vadi2:
Add handling for SSH-style Git URLs like git@github.com:org/repo.git in getURLWithNoUserInfo(). These URLs are not valid according to java.net.URL parsing, but are safe to return as-is since they don't contain embedded credentials (they use SSH keys).

The fix adds a regex check to detect SSH URL format before attempting standard URL parsing.


grahamegrieve (approved):
ok

PR #1209, opened 2025-12-12
2025 12 gg minor fixes

grahamegrieve:


org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/PublisherGenerator.java
```diff
@@ -539,6 +540,20 @@ public void generate() throws Exception {
     }
   }
 
+  private byte[] fixLang(byte[] b, String l) {
```
dotasek: Might be better to name this fillLangTemplate, unless this is actually fixing some incorrect content.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/web/PublicationProcess.java
```diff
@@ -615,7 +615,7 @@ private void doPublish(File fSource, File fOutput, JsonObject qa, String destina
               if (!v.cibuild() && !v.current()) {
                 String pv = v.path();
                 String vCode = pv.substring(pv.lastIndexOf("/")+1);
-                String dv = Utilities.path(destination, vCode);
+                String dv = Utilities.path(fRoot, relPath);
```
dotasek: I feel like dv might actually be 'path' or some other combo of these path manipulations already, but I can't quite hit it in debug traces.

org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/PublisherGenerator.java
grahamegrieve: ok

PR #1210, opened 2025-12-15
Br updates 2025 12

brynrhodes:
* Fixed npm model info loader not using the model info reader factory
* AdjunctFileLoader now supports CQL modelinfo files
* Corrected AdjunctFileLoader support XML format output setting in cql-options
* Fixed ELM data requirements incorrectly resolving code system references in value set, code, and concept declarations
* Added support for setting whether or not to correct model URLs
* Added logic to fixup references to implicitly loaded ModelInfo so the correct dependencies are reported
* Added effective data requirements processing for Measure, ActivityDefinition, PlanDefinition, and Questionnaire resources

### DIFF
--- a/org.hl7.fhir.publisher.core/pom.xml
+++ b/org.hl7.fhir.publisher.core/pom.xml
@@ -48,6 +48,11 @@
         </dependency>
         <dependency>
             <groupId>info.cqframework</groupId>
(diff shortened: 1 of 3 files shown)


grahamegrieve (approved):
ok

PR #1216, opened 2025-12-17
Update jena-shex version to 5.6.0

dotasek:


dotasek:
Since this is a dependency update, I have also run the QA Delta action: https://github.com/HL7/fhir-ig-publisher/actions/runs/20317959198

There should be no unexpected diffs or exceptions for that run.

PR #1249, opened 2026-02-04
Use specialised renderers for logical model instances

johngrimes:
When rendering logical model instances (Binary resources containing logical model data), the publisher now attempts to use specialised renderers from `RendererFactory` instead of always displaying raw JSON/XML.

For example, a `ViewDefinition` instance will now render using `ViewDefinitionRenderer` if available, producing a more readable narrative than syntax-highlighted JSON.

This restores functionality previously worked, but a regression broke it at some point.

**Changes:**
- Query `RendererFactory` for a specialised renderer matching the logical model type
- If one exists (i.e., not the generic `ProfileDrivenRenderer`), use it to build the narrative
- Fall back to JSON/XML rendering if no specialised renderer exists or if rendering fails
- Add null safety checks for `contentType` to prevent NPEs

**Fallback behaviour:** If the specialised renderer throws an exception or returns null, the code gracefully falls back to the existing JSON/XML rendering, ensuring no regression for edge cases.

### DIFF
--- a/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/PublisherGenerator.java
+++ b/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/PublisherGenerator.java
@@ -1117,32 +1117,46 @@ public class PublisherGenerator extends PublisherBase {
         }
         String html = null;
         if (rX.getLogicalElement() != null) {
-          String rXContentType = rX.getElement().getNamedChildValueSingle("contentType");
-          if (rXContentType.contains("xml")) {
-            org.hl7.fhir.r5.elementmodel.XmlParser xmlParser = new org.hl7.fhir.r5.elementmodel.XmlParser(this.pf.context);
-            XmlXHtmlRenderer xmlXHtmlRenderer = new XmlXHtmlRenderer();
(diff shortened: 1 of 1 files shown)


johngrimes:
Thanks @lmckenzi for taking a look at this, and apologies for the delay in getting back to you.

This change is confined to fixing the invocation of three specialised renderers for logical models. These specialised renderers were pre-existing, and were broken at some point:

- ViewDefinition
- WebTemplate
- FeatureDefinition

My understanding is that the rendering of logical models in IGs does not depend on versioning.

It is run from a switch statement off the type name in [RendererFactory in FHIR core](https://github.com/hapifhir/org.hl7.fhir.core/blob/8298a6d28dbd42668bd86da580df5f6db4fe30e7/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/renderers/RendererFactory.java).

The version doesn't seem to come into it at all.

For some reason something was changed here in the IG builder repo that broke this - this is an attempt to restore the functionality.

When taking another look at this, I noticed an area of the change that could be simplified further - so I committed this in [a5790b3](https://github.com/HL7/fhir-ig-publisher/pull/1249/commits/a5790b30153cd44970642fec9e886fae8ecda21a).

grahamegrieve (approved):
ok

PR #1253, opened 2026-02-20
Do not warn about unresolved CQL Library reference if the canonical is rooted in example.org

brynrhodes:


richardtownleyoneill-ADHA:
There are other example URLs: example.com, example.net. See https://www.iana.org/help/example-domains

PR #1256, opened 2026-02-22
perf: Batch SQLite inserts in DBBuilder — 85s → 0.1s in htmlOutputs phase

jmandel:
## Summary

During HTML output generation, `DBBuilder` writes ValueSet expansion codes and CodeSystem concept data into `package.db` (SQLite). The existing code calls `executeUpdate()` for every individual row with autocommit enabled. Each call triggers a disk fsync, turning a trivial data load into an I/O storm that dominates the entire `htmlOutputs` phase.

This PR wraps all insert loops in explicit transactions and uses JDBC batch operations (`addBatch()` / `executeBatch()`), reducing ~20,000 individual autocommitted inserts to a single transaction with one fsync.

## What changed

**`DBBuilder.java`** — single file, no API changes to callers:

1. **`recordExpansion()`**: Now queues expansions into a `ConcurrentLinkedQueue` instead of writing inline during parallel rendering. This eliminates the `synchronized` lock contention that serialized all rendering threads on DB writes.

2. **`flushDeferredOperations()`** (new): Drains the queue after the parallel phase completes, writing all ValueSet expansion codes in a single transaction using `addBatch()` + `executeBatch()` with periodic flush every 5,000 rows.

3. **`finishResources()`**: Wraps all 5 insert loops (Properties, Concepts, ConceptProperties, Designations, ConceptMappings) in a single transaction with `addBatch()`/`executeBatch()`. Includes proper rollback on error.

4. **`addConcepts()` / `addConceptProperties()` / `addConceptDesignations()`**: Changed from `executeUpdate()` to `addBatch()` (called within the transaction from `finishResources`).

## Why it was slow

SQLite with autocommit wraps each `executeUpdate()` in its own transaction. Each transaction commit forces a disk fsync to ensure durability. For ~20,000 ValueSet codes:

- **Before**: 20,000 transactions × 1 fsync each = 20,000 fsyncs ≈ 85 seconds
- **After**: 1 transaction × 1 fsync = 0.1 seconds

The `recordExpansion()` method was also `synchronized`, meaning all rendering threads serialized on this lock to do individual DB writes — completely defeating multi-threaded HTML generation.

### htmlOutputs phase
| Metric | Before | After (1T) | After (12T) |
|--------|--------|-----------|------------|
| htmlOutputs wall time | 100.3s | **16.7s** | **5.8s** |
| DB flush time | 84.7s | 0.1s | 0.1s |
| DB cumulative time | 85.0s | 2.2s | 2.2s |

### Overall build
| Metric | Before | After (12T) | Change |
|--------|--------|-------------|--------|
| Total wall time | 282s | **186s** | **−34%** |

### Multi-thread scaling (now unblocked)
| Threads | Before | After |
|---------|--------|-------|
| 1 thread | 103s | 16.7s |
| 12 threads | 94s (−9%) | 5.8s (−65%) |

Before this fix, adding threads only improved the ~16s rendering portion while the 85s single-threaded DB flush dominated. After the fix, 12 threads achieve 3× speedup on rendering.

### Output correctness
All `package.db` table row counts identical between before/after builds:
- Resources: 350, ConceptMappings: 673, ValueSet_Codes: 19,932
- Plus 12 other tables, all matching

HTML output diff shows only expected non-deterministic differences (UUIDs, xlsx timestamps).

## Why this was not caught before

1. **CPU profilers miss I/O wait**: fsync time appears as idle/blocked, not CPU. Profilers attributed the time to the rendering call stack that invoked `recordExpansion()`.
2. **Aggregated DB timing**: The publisher reports "DB Cumulative Time" at build end but doesn't break it down by phase, hiding the 85s inside htmlOutputs.
3. **Single-thread testing masks the issue**: With 1 thread the DB writes interleave with rendering, making total time look like "rendering is slow" rather than "DB is slow".

## Risk assessment

**Low risk**:
- Single file changed, no API changes
- Pure performance optimization — identical output
- Standard JDBC batch pattern, well-understood
- Transaction rollback on error preserves data integrity
- Periodic `executeBatch()` every 5,000 rows limits memory usage
- No changes to table schema or query patterns

### DIFF
--- a/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/renderers/DBBuilder.java
+++ b/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/renderers/DBBuilder.java
@@ -267,6 +267,18 @@ public class DBBuilder {
 
   private long cumulativeTime;
   
+  // Deferred DB operations: collected during parallel phases, flushed after
+  private final java.util.concurrent.ConcurrentLinkedQueue<DeferredExpansion> deferredExpansions = new java.util.concurrent.ConcurrentLinkedQueue<>();
+  
+  private static class DeferredExpansion {
+    final ValueSet vs;
+    final ValueSetExpansionOutcome exp;
+    DeferredExpansion(ValueSet vs, ValueSetExpansionOutcome exp) {
+      this.vs = vs;
+      this.exp = exp;
+    }
+  }
+  
   private void time(long start) {
     cumulativeTime = cumulativeTime + (System.currentTimeMillis() - start);
   }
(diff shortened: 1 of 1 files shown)


grahamegrieve:
why a draft?

PR #1270, opened 2026-03-13
Add OAuth client_credentials support for terminology server authentication

dionmcm:
## Purpose
The IG Publisher currently only supports authentication with terminology servers via a static OAuth token passed into the process. Users whose security policies enforce short token expiry (e.g. 30 minutes) find that the token expires before the publisher finishes, causing build failures. This PR addresses that by allowing configuration of OAuth 2.0 client_credentials flow directly, so the publisher can automatically acquire and refresh tokens as needed.

Depends on https://github.com/hapifhir/org.hl7.fhir.core/pull/2348 in org.hl7.fhir.core, which the IG Publisher uses to perform FHIR operations on the terminology server.

## Summary
- Adds CLI parameters (`-tx-client-id`, `-tx-client-secret`, `-tx-token-endpoint`) to authenticate with FHIR terminology servers using OAuth 2.0 client_credentials grant
- Includes input validation (missing params, whitespace-only values, flag-as-value detection, URL syntax/scheme checks) with clear error messages
- Normalizes TX server URL via `makeSecureRef()` ~~and supports `-devtx` fallback when `-tx` is absent~~
- Updates `removePassword()` in both `Publisher` and `PublicationProcess` to mask `-tx-client-secret` values in log output

## Test plan
- [x] 14 integration tests covering: CLI arg parsing, full end-to-end token acquisition flow with MockWebServer, password masking, missing/blank/whitespace param validation, flag-as-value detection, no-op when OAuth args absent, default/devtx TX fallback, malformed URL rejection, non-HTTP scheme rejection
- [x]  Manual: run publisher with `-tx-client-id`, `-tx-client-secret`, `-tx-token-endpoint` against an OAuth-protected TX server

### DIFF
--- a/org.hl7.fhir.publisher.core/pom.xml
+++ b/org.hl7.fhir.publisher.core/pom.xml
@@ -149,6 +149,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>mockwebserver</artifactId>
+            <version>${okhttp.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.apache.santuario</groupId>
             <artifactId>xmlsec</artifactId>
             <version>4.0.4</version>
--- a/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java
+++ b/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/Publisher.java
@@ -1670,35 +1670,75 @@ public class Publisher extends PublisherBase implements IReferenceResolver, IVal
     }
   }
 
-  private static String removePassword(String[] args, int i) {
+  static String removePassword(String[] args, int i) {
     if (i == 0) {
       return args[i];
     }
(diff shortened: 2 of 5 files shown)


dotasek (changes requested):
This PR requires changes included in https://github.com/hapifhir/org.hl7.fhir.core/pull/2348/

However, that PR is not in a mergable state. Until that is resolved, this PR can't be approved either.

grahamegrieve:
@dionmcm can you revisit this now that all the http work is landed?

dionmcm:
@grahamegrieve I've reviewed and changed the https://github.com/hapifhir/org.hl7.fhir.core/pull/2348 and I think I should just close and abandon this one.

The main point to this PR was to provide CLI parameters, mainly thinking of CI pipelines. But `fhir-settings.json` can be used with the changes I've proposed to org.hl7.fhir.core to use client credentials without the CLI parameters. To simplify things I've removed them from the org.hl7.fhir.core PR, and therefore this PR can be dropped.

If there is desire for the CLI parameters, I'm happy to create a new pair of PRs that add that ability. But I thought it was better to keep it minimalist for the moment, particularly with all the org.hl7.fhir.core changes going on I thought that would be easier to review and merge for @dotasek.

PR #1294, opened 2026-04-18
Fix NPE in PublicationChecker when publication-request.json is absent

costateixeira:
When an IG has been previously published (canonical serves a valid package-list.json) but the local repo has no publication-request.json, checkExistingPublication NPE'd on pr.has("movedFrom"). 

This treats pr == null as "no movedFrom declared" and fall through to the normal publication-comparison branch.


grahamegrieve (approved):
ok

PR #1305, opened 2026-05-17
Add support for rendering FML fragments in IGs

brianpos:
Add support for FML fragments in IGs.
Which can be used in a markdown file like this:
```
```fml
{% fragment StructureMap/MetadataBasic fml %}
` ``
```


grahamegrieve (approved):
ok

PR #1315, opened 2026-05-30
fix show-inherited-invariants:false hiding a profile's own invariants

### LINKED ISSUE #321: Own invariants are not shown when using show-inherited-invariants: false
See https://chat.fhir.org/#narrow/stream/196008-ig-publishing.20requirements/topic/Invariants - may be superseeeded by https://github.com/HL7/fhir-ig-publisher/issues/272

vadi2:
The Constraints-table filter compared constraint.source to the profile url with exact string equality, but pin-canonicals appends |version to source in the render model, so own invariants were misread as inherited and dropped. Compare version-insensitively, and drop the genMode!=DIFF escape so the snapshot/key tables also honour the flag. Default behaviour is unchanged.

Confirmed fixed:

image

Fixed #321


jkiddo (approved):

grahamegrieve (approved):
ok. thanks

grahamegrieve (approved):
ok

PR #1316, opened 2026-05-30
Added support for ImplementationGuide:base, to allow differentiation …

lmckenzi:
…of the 'primary' ImplementationGuide from any example IGs.  Also fixed the logic so that example templates actually function rather than auto-defaulting to 'ANY' if there isn't a specific resource template.


grahamegrieve (approved):
ok

PR #1327, opened 2026-06-17
Fix cloud/static-hosting redirects: emit real HTML and wire CLOUD into go-publish

KyleOps:
## Problem

`ServerType.CLOUD` (`"server": "cloud"` in `publish-setup.json`) is intended for static
hosts — S3, GitHub Pages, Netlify, etc. that cannot execute server-side redirect
scripts. Two defects make the CLOUD path non-functional, so a cloud-hosted IG cannot get
working canonical-URL redirects.

### Bug 1: `HTML_TEMPLATE` is a copy of `PHP_TEMPLATE`
`IGReleaseRedirectionBuilder.HTML_TEMPLATE` is a verbatim duplicate of `PHP_TEMPLATE`,
so `createHtmlRedirect()` writes PHP source into `index.html`. A static host serves that
as `text/html`, so the browser renders the raw PHP (including the trailing
"…PHP has failed badly" text) and never redirects.

### Bug 2: `-go-publish` never selects the CLOUD path
`PublicationProcess.updatePublishBox` (the `-go-publish` path) has no `CLOUD` branch, so
cloud sites fall through to `else if (!canonical.contains("hl7.org/fhir"))` →
`buildApacheRedirections()` and get PHP regardless of the configured server type.
(`IGReleaseUpdater` already handles `CLOUD`; this brings the go-publish path into line.)

## Fix

1. Replace `HTML_TEMPLATE` with a real static HTML redirect: `<meta http-equiv="refresh">`
   + `window.location.replace(...)` + `<link rel="canonical">`, plus JSON/XML links for
   machine clients. Content negotiation on `Accept` is not possible without a server
   runtime, so browsers are sent to the human-readable page.
2. Add the missing `else if (serverType == ServerType.CLOUD) rb.buildCloudRedirections();`
   case to `PublicationProcess.updatePublishBox`, matching `IGReleaseUpdater`.

## Testing

`mvn -pl org.hl7.fhir.publisher.core compile` passes. Generated `index.html` verified to
redirect correctly for ImplementationGuide / StructureDefinition / CodeSystem / ValueSet
canonical paths.

### DIFF
--- a/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/web/IGReleaseRedirectionBuilder.java
+++ b/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/web/IGReleaseRedirectionBuilder.java
@@ -91,31 +91,24 @@ public class IGReleaseRedirectionBuilder {
       "    \r\n"+
       "You should not be seeing this page. If you do, PHP has failed badly.\r\n";
 
-  private static final String HTML_TEMPLATE = "<?php\r\n"+
-      "function Redirect($url)\r\n"+
-      "{\r\n"+
-      "  header('Location: ' . $url, true, 302);\r\n"+
-      "  exit();\r\n"+
-      "}\r\n"+
-      "\r\n"+
-      "$accept = $_SERVER['HTTP_ACCEPT'];\r\n"+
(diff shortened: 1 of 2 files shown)


grahamegrieve (approved):
ok

PR #1341, opened 2026-07-07
Make sure binary loading works on contained resources too, adds SVG s…

lmckenzi:
…upport and fixes JPEG support.


grahamegrieve (approved):
ok

PR #1351, opened 2026-08-03
Support .sql adjunct files as application/sql (#1323)

### LINKED ISSUE #1323: Support .sql files (application/sql) for ig-loader content attachments
When authoring SQL on FHIR SQLView and SQLQuery `Library` resources, it would be very helpful to attach the SQL from a separate file using the existing `ig-loader-` mechanism, the same way CQL content IGs attach `.cql` files. That keeps the SQL editable and reviewable as a plain file rather than hand-encoded base64 in the resource.

At the moment this does not quite work for SQL. `AdjunctFileLoader.determineContentType()` recognises `pdf`, `dicom`, `png`, `gif`, `jpeg`, `cql`, `feature`, `json`, `xml`, `text` and `txt`, but not `sql`, so it returns `null` and `content.contentType` is never set. The SQL on FHIR SQLView/SQLQuery profiles require `content.contentType` to be present and to start with `application/sql` (their `sql-must-be-sql-expressions` invariant), so a loaded `.sql` attachment fails validation. And because `checkReplaceable()` only fires when the attachment has no `contentType` preset, it is not possible to set `application/sql` by hand and still use the loader.

Would you consider adding a mapping from the `sql` extension to `application/sql` in `determineContentType()`? That single addition would let SQLView and SQLQuery Libraries reference their SQL via `content.id = ig-loader-<file>.sql` and have the publisher base64-encode it with the correct content type, mirroring the CQL workflow and avoiding hand-maintained base64.

Reference: `org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/loaders/AdjunctFileLoader.java`, method `determineContentType()`.

johngrimes:
Fixes #1323.

`AdjunctFileLoader.determineContentType()` did not recognise the `sql` extension, so a `.sql` file loaded via `content.id = ig-loader-<file>.sql` produced an `Attachment` with no `contentType`.

The SQL on FHIR SQLView and SQLQuery profiles require `content.contentType` to be present and to start with `application/sql` (their `sql-must-be-sql-expressions` invariant), so such an attachment fails validation. And because `checkReplaceable()` only fires when the attachment has no `contentType` preset, it was not possible to set `application/sql` by hand and still use the loader.

## Change

Map the `sql` extension to `application/sql`, the IANA registered media type ([RFC 6922](https://www.rfc-editor.org/rfc/rfc6922.html)). This lets SQL on FHIR Libraries reference their SQL as a plain, reviewable file rather than hand-encoded base64, mirroring the existing `.cql` workflow.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,0 +1,1 @@
+* Loader: Support .sql adjunct files (application/sql) for ig-loader content attachments
--- a/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/loaders/AdjunctFileLoader.java
+++ b/org.hl7.fhir.publisher.core/src/main/java/org/hl7/fhir/igtools/publisher/loaders/AdjunctFileLoader.java
@@ -233,6 +233,8 @@ public class AdjunctFileLoader {
       return "text/cql";
     } else if ("feature".equals(ext)) {
       return "text/x-gherkin";
+    } else if ("sql".equals(ext)) {
+      return "application/sql";
     } else if ("json".equals(ext)) {
       return "application/json";
     } else if ("xml".equals(ext)) {


grahamegrieve (commented):
ok

PR #1353, opened 2026-08-05
Add -ssrf-protection-enabled param + explicit method for settings

dotasek:
Adds an equivalent to the validator `-ssrf-protection-enabled` param to disable ssrf protection via the CLI.

This has been refactored to a single method that ensures the correct order of loading fhir-settings.json and then using the CLI as an override.


grahamegrieve (approved):
ok

